PR #2, opened 2025-04-01
Add create and teardown cluster playbooks

adriengentil:
This PR creates 2 playbooks to create and delete hosted cluster. To perform this action,I created 3 roles:
- `cluster_working_namespace`: find and return the namespace name in which all the resources will be created
- `cluster_settings`: contains the logic to compute the settings which will be used to create the hosted cluster, it basically merges settings set at template level the ones provided by the user
- `hosted_cluster`: creates the resources in the namespace returned by `cluster_working_namespace` and use the settings returned `cluster_settings` to customize the deployment

## Summary by CodeRabbit

- **Documentation**
  - Added a new "Pre-requisites" section with clear setup instructions and dependency installation steps.

- **New Features**
  - Enhanced automation workflows for creating and tearing down hosted control plane clusters using an updated, order-driven approach.
  - Improved hosted cluster management with dynamic configuration of secrets, settings, node pools, and refined namespace detection.
  - Introduced a new YAML configuration file for defining cluster orders and default settings.
  - Added support for managing Kubernetes resources related to hosted clusters.

- **Chores**
  - Updated project configuration to enforce Python 3.13 and manage dependency requirements.
  - Refined tool settings and internal defaults for improved consistency.


roles/hosted_cluster/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+          import.open-cluster-management.io/hosting-cluster-name: "{{ hosted_cluster_name }}"
+          import.open-cluster-management.io/klusterlet-deploy-mode: Hosted
+          open-cluster-management/created-via: hypershift
+        labels:
+          name: "{{ hosted_cluster_name }}"
+          cloud: BareMetal
+          vendor: OpenShift
+          cluster.open-cluster-management.io/clusterset: default
+      spec:
+        hubAcceptsClient: true
```
larsks, line 127: Do we to create this in the playbook? In the past when I've deployed a hostedcluster, I haven't needed to create the ManagedCluster resource explicitly.

roles/hosted_cluster/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+          name: "sshkey-cluster-{{ hosted_cluster_name }}"
+        networking:
+          clusterNetwork:
+            - cidr: 10.132.0.0/14
+          serviceNetwork:
+            - cidr: 172.31.0.0/16
+        platform:
+          type: Agent
+          agent:
+            agentNamespace: ''
```
larsks, line 57: Doesn't this need to be set to the namespace that contains the target infraenv?

roles/hosted_cluster/tasks/main.yml
```diff
@@ -0,0 +1,127 @@
...
+        replicas: "{{ hosted_cluster_template.parameters.number_of_nodes }}"
+        management:
+          autoRepair: false
+          upgradeType: InPlace
+        platform:
+          type: Agent
+          agent:
+            agentLabelSelector:
+              matchLabels:
+                infraenvs.agent-install.openshift.io: "{{ hosted_cluster_template.infraenv }}"
```
larsks, line 104: Flagging this for discussion, because I think it's odd that we have ended up with `infraenv` as a template parameter. I think -- at least for POC1, and possibly going forward -- that we will also want to match agents on cluster name, and some other component will be responsible for assigning agents to a cluster (e.g., we would do that by hand for POC1).

roles/hosted_cluster/tasks/main.yml
adriengentil: Not sure, I basically copy-pasted what the OCM UI does when creating a hosted cluster. I'm happy to remove it if it's not needed.

roles/hosted_cluster/tasks/main.yml
adriengentil: I agree that something more functional would be great.

About the var file used as a template, I was thinking about the  `parameters` section to be merged with the parameters provided by the user in the cluster order, so everything outside this section would be hard-coded by the template.

roles/hosted_cluster/tasks/main.yml
adriengentil: I see this manifest mentioned here in this blog post: https://www.redhat.com/en/blog/how-to-build-bare-metal-hosted-clusters-on-red-hat-advanced-cluster-management-for-kubernetes

.python-version
```diff
@@ -0,0 +1 @@
+3.13
```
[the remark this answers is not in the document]
adriengentil: Python 3.13.0 was released stable on Oct. 7, 2024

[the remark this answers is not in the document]
mhrivnak: RHEL 9 only goes up to 3.12 though. And its default is still 3.9.

mhrivnak (approved):

PR #5, opened 2025-04-07
Fix cloudkit label name and label all created resources

adriengentil:


mhrivnak (approved):
The rabbit may have some valid feedback, but otherwise this LGTM.

PR #10, opened 2025-04-08
enable parallel processing of events in EDA

adriengentil:
We currently observe that client are getting timeouts from the EDA
webhooks. One reason of this is might be because the processing of the
events is sequencial by default.

This PR enables parallel processing of event in EDA, which should fix the
client timeouts we currently experience.

Doc ref: https://ansible.readthedocs.io/projects/rulebook/en/latest/rulebooks.html#rulesets

### DIFF
--- a/rulebooks/cluster_fulfillment.yml
+++ b/rulebooks/cluster_fulfillment.yml
@@ -1,6 +1,7 @@
 ---
 - name: Fulfillment webhook
   hosts: localhost
+  execution_strategy: parallel
   sources:
     - ansible.eda.webhook:
         host: 0.0.0.0


mhrivnak:
After tomorrow's demo, it would be great to understand the real cause for the timeout. I wouldn't normally expect a system like EDA to just not respond when its queue grows.

adriengentil:
Here is my understanding:

The ruleset engine in EDA shares a bounded queue of `1` with the source plugin (webhook is our case):
 https://github.com/ansible/ansible-rulebook/blob/50f1ffdd01b0e3f0f1de20a10693d7ddf81b7a27/ansible_rulebook/app.py#L259C1-L259C40

Then, items are de-queued in the rule engine (https://github.com/ansible/ansible-rulebook/blob/50f1ffdd01b0e3f0f1de20a10693d7ddf81b7a27/ansible_rulebook/rule_set_runner.py#L318), we'll wait for the completion of the playbook before starting the processing the next rule when we are in sequential mode. That means the webhook plugin needs to wait for the queue to be empty in order to insert the current event and return a response to the client. And it think that's why we get a timeout, because the playbook execution duration are quite long vs the number of events triggered by the controller.

larsks (approved):

computate:
@larsks @adriengentil I don't know if these are relevant, but there are some Ansible EDA control variables that might help with parallel processing: 

| RPM variable name | Container variable name|Description |
|---------------------------- | --------------------------------- | ---------------- |
| automationedacontroller_activation_workers | eda_activation_workers   | Number of workers used for ansible-rulebook activation pods in Event-Driven Ansible. |
| automationedacontroller_default_workers |  eda_workers | Number of workers used in Event-Driven Ansible for application work.

PR #16, opened 2025-04-09
Add config for building execution environment

larsks:
Tools and configuration to run playbooks that interact with both OpenStack/ESI and OpenShift.

To build the execution environment

1. Download the `openshift-clients` package from [the customer portal] and place it in
    `openshift-clients.rpm`

2. Run:

    ```
    ansible-builder build --tag cloudkit-aap-ee
    ```

## Summary by CodeRabbit

- **Documentation**
	- Introduced a README outlining setup instructions and necessary tools for running the execution environment.
- **Chores**
	- Added configuration files and dependency lists to streamline the build process and ensure the required packages and collections are available.
	- Updated the `.gitignore` to exclude the `context/` directory.
	- Added new requirements for Python packages in the execution environment.
	- Removed the explicit collection dependency for `kubernetes.core`.


execution-environment/execution-environment.yaml
```diff
@@ -0,0 +1,26 @@
...
+  ansible_core:
+    package_pip: ansible-core
+  ansible_runner:
+    package_pip: ansible-runner
+
+options:
+  package_manager_path: /usr/bin/microdnf
+
+additional_build_files:
+  - src: openshift-clients.rpm
```
adriengentil, line 19: An alternative would be to get the clients as a tar from https://mirror.openshift.com/pub/openshift-v4/x86_64/clients/ocp/stable/ which is publicly available.

execution-environment/execution-environment.yaml
```diff
@@ -0,0 +1,26 @@
+version: 3
+
+images:
+  base_image:
+    name: registry.redhat.io/ansible-automation-platform/ee-minimal-rhel9:2.18
```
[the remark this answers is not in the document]
larsks: I was hoping for something with a version tag for which we could get minor updates for free; 2.18 is the most recent version-tagged image. I guess maybe they've stopped using version tags?

execution-environment/requirements.yaml
```diff
@@ -0,0 +1,11 @@
+collections:
```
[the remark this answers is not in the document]
larsks: Do we even need both files? Or since we're using a custom execution environment with everything pre-installed, can we drop ../collections?

execution-environment/execution-environment.yaml
larsks: Good plan. I'll make that change.

larsks: Wait, no, we can't -- at least, not without getting hacky. There is an rpm dependency on the `openshift-clients` package, and the environment will fail to build without it.

larsks: The build fails like this:

```
+ PKGMGR_OPTS='--nodocs --setopt install_weak_deps=0'
+ mkdir -p /output/bindep
+ mkdir -p /output/wheels
+ mkdir -p /tmp/src
+ cd /tmp/src
+ PACKAGES=
+ PIP_OPTS=
+ install_bindep
+ '[' -f bindep.txt ']'
+ bindep -l newline
+ sort
+ '[' rhel == centos ']'
++ bindep -b compile
++ true
+ compile_packages=openshift-clients
+ '[' '!' -z openshift-clients ']'
+ /usr/bin/microdnf install -y --nodocs --setopt install_weak_deps=0 openshift-clients
error: No package matches 'openshift-clients'
Error: building at STEP "RUN /output/scripts/assemble": while running runtime: exit status 1
```

Let me see if there's any way to make this optional.

adriengentil: what is that dependency?

execution-environment/requirements.txt
```diff
@@ -0,0 +1,5 @@
+python-openstackclient
```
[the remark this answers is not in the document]
larsks: Were you thinking of having this automated somehow?

execution-environment/execution-environment.yaml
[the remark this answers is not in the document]
adriengentil: I guess the logic is "here is the container to use for aap 2.5, you don't need to care about rhel minor version"

execution-environment/execution-environment.yaml
larsks: I'm not sure, but there's some discussion about it here: https://access.redhat.com/solutions/6985157

larsks: Ah, that says:

> The collection [redhat.openshift](https://console.redhat.com/ansible/automation-hub/repo/published/redhat/openshift/) has a dependency on the collection [kubernetes.core](https://console.redhat.com/ansible/automation-hub/repo/published/redhat/openshift/dependencies/). One of the requirements listed in the [bindep.txt](https://access.redhat.com/bounce/?externalURL=https%3A%2F%2Fgithub.com%2Fansible-collections%2Fkubernetes.core%2Fblob%2Fmain%2Fbindep.txt) file for installing the kubernetes.core collection is the package openshift-clients, which is a component of the RHOCP repositories.

execution-environment/requirements.txt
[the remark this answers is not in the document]
adriengentil: yes, I think it would be great to re-compile a requirement.txt when we change the dependencies, it can be a `pre-commit` hook. But I'm fine to do it as part of another PR.

execution-environment/execution-environment.yaml
adriengentil: otherwise  with image has the openshift-clients pre-installed: https://catalog.redhat.com/software/containers/ansible-automation-platform-25/ee-supported-rhel9/650a56807ab682340da38ad5?container-tabs=packages&rh-tabs-0yyn0ehxp=rh-tab-tg1692g4p

but it don't know if it makes sense to use it with the ansible builder

adriengentil: feel free to choose whatever solution, it looks like there no easy way :cry:

larsks: I think I'll just leave it as is for now. In the future, maybe we'll add some credentials to a github workflow so that it could download the package and build the image.

larsks:
@adriengentil I think this is ready to go.

adriengentil (approved):

Issue #18, opened 2025-04-11
Organize roles into collections


adriengentil:
Since we will write roles that have different responsibilities, it would be great to split them is several folders. Since Ansible expects `roles/` folder to be organized in a flat structure, I think the best way would be to put them into different collections.

PR #22, opened 2025-04-14
Refactor roles

larsks:
@adriengentil what do you think about this as a first step towards cleaning things up?

This moves ESI-specific functionality into a local collection (`massopencloud/esi`), and refactors the roles so that we no longer have nested roles.

Addresses #18


adriengentil:
Yep, I was experimenting something like that yesterday in one on my branches (https://github.com/adriengentil/cloudkit-aap/tree/move-to-collections), while it just works locally with `ansible-playbook`, I have issues to run it in AAP, it doesn't looks at the roles into the collection.

.gitignore
```diff
@@ -8,6 +8,7 @@ collections/ansible_collections/*
 !collections/ansible_collections/cloudkit
 collections/ansible_collections/cloudkit/*
 !collections/ansible_collections/cloudkit/global_config
```
adriengentil, line 10: I think those things are left overs from the time I did create the repo, I think we can delete

collections/ansible_collections/massopencloud/esi/roles/floating_ip/tasks/main.yaml
```diff
@@ -0,0 +1,17 @@
+- name: Validate floating_ip_state parameter
```
adriengentil, line 1: validation can be done through argument specs https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-argument-validation

larsks: Oh, neat; I've updated the role to use argument validation.

tzumainn (commented):
> Is the plan to refactor the port forwarding into the massopencloud.esi collection at some point as well?

Yes. Ultimately, we want to move all the interaction with ESI into collections so that our primary playbooks don't really need to know anything about ESI. But I think we can do more of that in future PRs. I'd like to get this merged so we can build on top of it.

tzumainn (approved):
Okay! I don't know about the naming issues lint is pointing out, but the ESI bits look good to me!

Issue #8, opened 2025-04-07
Cluster templates should be roles


larsks:
It looks like right a now a "cluster template" is just a variable file, but that's not going to work in the long term -- we need cluster templates to be able to perform arbitrary post-install tasks (like, "install openshift ai" or "configure oauth", etc). A "cluster template" should be a role that then includes other roles in order to get things done. E.g., if I ask for an `ocp_4_17_small` cluster, that should have a tasks file that looks something like:

```
- name: Install the hosted cluster
  include_role:
    name: hosted_cluster
```

Whereas a more complicated template might look like:

```
- name: Install the hosted cluster
  include_role:
    name: hosted_cluster

- name: Install and configure OpenShift AI
  include_role:
    name: install_openshift_ai

- name: Configure github oauth
  include_role:
    name: github_oauth
```

jhernand:
@larsks where will that `meta/cloudkit.yaml` file be stored? Is it part of AAP? Is there an API to fetch it?

PR #24, opened 2025-04-16
Propose a template architecture

larsks:
This commit proposes how we might structure cluster templates:

1. Templates live in the `cloudkit.templates` ansible collection.

2. Metadata specific to our solution lives in `meta/cloudkit.yaml` (this includes the template display name, description, and information about node requests).

3. Template parameters are documented using Ansible argument validation [1]. There are some example parameters provided in this template; they are meant strictly to demonstrate the structure of the `argument_specs.yaml` file and are not meant to suggest features we need to implement for POC3.

There is an open question about how exactly these template roles will fit into our playbook logic. I plan to address that in a subsequent commit.

[1]: https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-argument-validation

Addresses: innabox/issues#98, #8


larsks:
If we refer to fully qualified template names in the API (`cloudkit.templates.ocp_4_18_small`), this also suggests a convenient mechanism for a datacener administrator to provide their own custom templates (by having their own `mycompany.templates` collection).

This requires a minor changes to the API and the CRD to accept `.` in template ids, but I think it makes sense.

tzumainn:
I don't see an obvious way to do this, but it would be nice if `argument_specs` could be imported from elsewhere, allowing individual templates to import from a common "library" of arguments; otherwise, it feels like there'll be a lot of duplication. Maybe that's not a concern?

tzumainn (approved):
Makes sense!

larsks:
There were some problem with the previous proposal; in particular, Ansible
does not allow us to provide all module arguments from a dictionary
variable, and this made it difficult to effectively make use of the
`templateParameters` dictionary in the cluster order.

With this latest commit, template modules now have a `template_parameters` option,
which takes a dictionary value. We can pass the value of
`templateParameters` from the ClusterOrder directly to this variable.

By locating template parameters in an option like this, we also have the
ability to specify role arguments that aren't provided via template
parameters, such as the node requests.

There are two example templates provided in this commit to show how this
might look.

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small/meta/argument_specs.yaml
```diff
@@ -0,0 +1,34 @@
+argument_specs:
+  main:
+    options:
+      cluster_order:
+        type: dict
+        required: true
```
adriengentil, line 6: do you think it's still worth to pass the `cluster_order`?

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small_github/tasks/create_template.yaml
```diff
@@ -0,0 +1,70 @@
+- name: Create cluster
+  ansible.builtin.import_role:
```
adriengentil, line 2: we could even include `cloudkit.templates.ocp_4_17_small` here

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small/tasks/delete_template.yaml
```diff
@@ -0,0 +1,19 @@
+- ansible.builtin.import_role:
+    name: hosted_cluster
```
adriengentil, line 2: I find it confusing to reference a top-level role from a collection, we should probably move these roles in a dedicated collection (or it the template collection directly?).

computate (approved):
This is a very useful template architecture based on AAP templates. This supports Ansible roles like @IsaiahStapleton is building for AI automated cluster configuration with Ansible Roles: 
https://github.com/IsaiahStapleton/ansible-rhoai

The options in the cloudkit.yaml for `default_node_request` and `allowed_resource_classes` are very useful for defining GPU and non-GPU cluster configurations.

schwesig (approved):

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small/tasks/delete_template.yaml
larsks: I don't think they should live in the template collection, but I can see moving them into a separate collection. Let's plan on that in a subsequent pr.

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small/meta/argument_specs.yaml
larsks: I think templates should have access to the cluster order. We're using it right now e.g. to access `spec.nodeRequests`.

mhrivnak:
Looks good to me. We can try this out and see how it goes. If we find opportunities to generalize some of the parameters, have a library, or find some common ones that would benefit from first-class inclusion in the ClusterOrder API, we can always add such things later.

PR #29, opened 2025-04-21
Add ESI network role

tzumainn:
* creating a network requires a name, and returns a uuid
* deleting a network requires either a name or a uuid; the openstack network role claims that it can filter by either name or uuid, but the latter does not seem to work

### DIFF
--- a/collections/ansible_collections/massopencloud/esi/roles/network/meta/argument_specs.yaml
+++ b/collections/ansible_collections/massopencloud/esi/roles/network/meta/argument_specs.yaml
@@ -1,0 +1,15 @@
+argument_specs:
+  main:
+    options:
+      network_state:
+        type: "str"
+        default: present
+        choices:
+        - present
+        - absent
+      network_name_or_id:
+        description: |
+          Name of the network when state is `present`.
+          Name or ID of network to be deleted when state is `absent`; the deletion will fail if the network name is not unique.
+        type: "str"
+        required: true
--- a/collections/ansible_collections/massopencloud/esi/roles/network/tasks/create_network.yaml
+++ b/collections/ansible_collections/massopencloud/esi/roles/network/tasks/create_network.yaml
@@ -1,0 +1,21 @@
+- name: Check if network exists
+  openstack.cloud.networks_info:
+    name: "{{ network_name_or_id }}"
+  register: networks
+
+- name: Set create_network_result to found network
+  when: networks.networks
+  ansible.builtin.set_fact:
+    create_network_result: "{{ networks.networks[0].id }}"
+
+- name: Create a new network
+  when: not networks.networks
+  block:
+  - name: Create network  # noqa:no-changed-when
+    openstack.cloud.network:
+      name: "{{ network_name_or_id }}"
+    register: network_cmd_result
+
+  - name: Set create_network_result to new network
+    ansible.builtin.set_fact:
+      create_network_result: "{{ network_cmd_result.network.id }}"
--- a/collections/ansible_collections/massopencloud/esi/roles/network/tasks/destroy_network.yaml
+++ b/collections/ansible_collections/massopencloud/esi/roles/network/tasks/destroy_network.yaml
@@ -1,0 +1,12 @@
+- name: Check that exactly one network matching the ident exists  # noqa:no-changed-when
+  ansible.builtin.command: >-
+    openstack network show {{ network_name_or_id }} -f json
+  register: network_show_cmd_raw
+
+- name: Unmarshal output from JSON
+  ansible.builtin.set_fact:
+    network_show_cmd_result: "{{ network_show_cmd_raw.stdout | from_json }}"
+
+- name: Destroy network  # noqa:no-changed-when
+  ansible.builtin.command: >-
+    openstack network delete {{ network_show_cmd_result.id }}
--- a/collections/ansible_collections/massopencloud/esi/roles/network/tasks/main.yaml
+++ b/collections/ansible_collections/massopencloud/esi/roles/network/tasks/main.yaml
@@ -1,0 +1,7 @@
+- name: Create a new network
+  when: network_state == 'present'
+  ansible.builtin.include_tasks: create_network.yaml
+
+- name: Destroy a network
+  when: network_state == 'absent'
+  ansible.builtin.include_tasks: destroy_network.yaml


tzumainn:
The network destroy command will need to also destroy subnets and routers, but that will be handled in the L3-related PRs.

collections/ansible_collections/massopencloud/esi/roles/network/tasks/destroy_network.yaml
```diff
@@ -0,0 +1,7 @@
+- name: Check that exactly one network matching the ident exists  # noqa:no-changed-when
+  ansible.builtin.command: >-
+    openstack network show {{ network_identifier }} -f json
```
adriengentil, line 3: is it required? We try to delete anyway, right?

collections/ansible_collections/massopencloud/esi/roles/network/tasks/destroy_network.yaml
```diff
@@ -0,0 +1,7 @@
...
+
+- name: Destroy network  # noqa:no-changed-when
+  ansible.builtin.command: >-
+    openstack network delete {{ network_identifier }}
```
adriengentil, line 7:
```suggestion
    openstack.cloud.network:
      name: "{{ network_name }}"
      state: absent 
```
looks like it's possible: https://docs.ansible.com/ansible/latest/collections/openstack/cloud/network_module.html#parameter-state

tzumainn: Oh, I tried it, based on that documentation; it doesn't actually seem to work if you use the UUID rather than the name, unfortunately.

collections/ansible_collections/massopencloud/esi/roles/network/tasks/destroy_network.yaml
tzumainn: OpenStack allows multiple networks with the same name, so I figure we want to throw an error if someone tries to delete a network by name and it matches multiple networks.

collections/ansible_collections/massopencloud/esi/roles/network/meta/argument_specs.yaml
```diff
@@ -0,0 +1,12 @@
...
+    options:
+      network_state:
+        type: "str"
+        required: true
+        choices:
+        - present
+        - absent
+      network_name:
+        type: "str"
+        required: true
```
adriengentil, line 12: `network_identifier` should be declared here, and I think `network_name` shouldn't be required because it's not used in case of deletion.

tzumainn: ah, good catch - thanks! should be fixed now

collections/ansible_collections/massopencloud/esi/roles/network/tasks/destroy_network.yaml
adriengentil: https://opendev.org/openstack/ansible-collections-openstack/src/branch/master/plugins/modules/network.py#L241

the module does actually find the network by name, before using its id later to perform the deletion.

tzumainn: I think it makes sense to require a name when creating a network; as it is right now, it will accept either a name or a UUID when deleting, as `show` accepts either. It seems like a reasonable compromise to me, but I'm open to other behaviors!

adriengentil: It's weird because the underlying function should take both id or name:
https://opendev.org/openstack/openstacksdk/src/branch/master/openstack/network/v2/_proxy.py#L2400-L2418

tzumainn: It is really strange - I double-checked just now, and can confirm that for whatever reason the openstack network module doesn't work with a ID. I've made the other changes; my reasoning for not grabbing the ID was that the command proves that there is only one matching network, but it is much safer to use the ID as you suggested. Let me know if you have any other questions or concerns!

collections/ansible_collections/massopencloud/esi/roles/network/tasks/create_network.yaml
```diff
@@ -0,0 +1,21 @@
+- name: Check if network exists
+  openstack.cloud.networks_info:
+    name: "{{ network_name }}"
```
adriengentil, line 3:
```suggestion
    name: "{{ network_name_or_id }}"
```

I'm actually thinking that is would be enough, if a user wants to create a network, they won't give an ID.

collections/ansible_collections/massopencloud/esi/roles/network/meta/argument_specs.yaml
```diff
@@ -0,0 +1,15 @@
+argument_specs:
+  main:
+    options:
+      network_state:
+        type: "str"
+        required: true
+        choices:
+        - present
+        - absent
+      network_name_or_id:
```
adriengentil, line 10:
```suggestion
      network_name_or_id:
      description: >
         Name of the network in case of network when state is `present`.
         Name or ID of the network when state is `absent`, the deletion will fail if the network name is not unique.
```

Add some description in case we go with `network_name_or_id` and remove the use of `network_name`

collections/ansible_collections/massopencloud/esi/roles/network/meta/argument_specs.yaml
```diff
@@ -0,0 +1,15 @@
+argument_specs:
+  main:
+    options:
+      network_state:
+        type: "str"
+        required: true
```
adriengentil, line 6:
```suggestion
        default: present
```
it's quite common to default to `present` in ansible world

adriengentil (commented):
What do you think about using `network_name_or_id` for creation as well? I think it would be clearer to use.

collections/ansible_collections/massopencloud/esi/roles/network/tasks/create_network.yaml
tzumainn: I do have some (minor) concerns that someone might supply an ID-like string and expect the ID to be set to that, but I'm guessing the benefits of having a consistent parameter name outweighs that!

collections/ansible_collections/massopencloud/esi/roles/network/meta/argument_specs.yaml
tzumainn: Done!

collections/ansible_collections/massopencloud/esi/roles/network/meta/argument_specs.yaml
tzumainn: Ah, thanks for the info - done!

adriengentil (approved):
Looks good :+1:

PR #30, opened 2025-04-24
Add ESI subnet role

tzumainn:
* creating a subnet requires a name and network identifier, and returns a uuid
* deleting a subnet requires either a name or uuid
* deleting a network will automatically delete associated subnets

### DIFF
--- a/collections/ansible_collections/massopencloud/esi/roles/subnet/defaults/main.yaml
+++ b/collections/ansible_collections/massopencloud/esi/roles/subnet/defaults/main.yaml
@@ -1,0 +1,6 @@
+subnet_cidr: 192.168.50.0/24
+subnet_gateway_ip: 192.168.50.254
+subnet_allocation_pool_start: 192.168.50.10
+subnet_allocation_pool_end: 192.168.50.200
+subnet_dns_nameservers:
+  - 8.8.8.8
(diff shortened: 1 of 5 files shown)


adriengentil (approved):
Nice !

collections/ansible_collections/massopencloud/esi/roles/subnet/defaults/main.yaml
```diff
@@ -0,0 +1,6 @@
+subnet_cidr: 192.168.50.0/24
```
larsks, line 1: We should default this to something larger than /24. I would probably go with a /22.

collections/ansible_collections/massopencloud/esi/roles/subnet/defaults/main.yaml
```diff
@@ -0,0 +1,6 @@
...
+subnet_gateway_ip: 192.168.50.254
```
larsks, line 2: I think it's much more common for this to be `.1`.

collections/ansible_collections/massopencloud/esi/roles/subnet/tasks/create_subnet.yaml
```diff
@@ -0,0 +1,27 @@
...
+
+- name: Set create_subnet_result to found subnet
+  when: subnets.subnets
+  ansible.builtin.set_fact:
+    create_subnet_result: "{{ subnets.subnets[0].id }}"
+
+- name: Create a new subnet
+  when: not subnets.subnets
+  block:
+  - name: Create subnet  # noqa:no-changed-when
```
larsks, line 14: Do we need the `noqa` annotations here? These are necessary when using `command` tasks, but we're not doing that here.

PR #36, opened 2025-04-30
Add agent management role

adriengentil:
New role that label the agents given the desired number of nodes and the resource class to be assigned the cluster.

The assumption I took is that we will have as one nodepools per group of nodes of the same resource class, so this role should be called for each nodepool (resource class) associated to the cluster being managed in the scope of this playbook/role.

In order for this role to cleanup the agents when the nodepool is down-scaled, it has to run after the nodepool has been updated, we then we watch which agents will be removed by hypershift, and clean them up.

https://github.com/issues/assigned?issue=innabox%7Cissues%7C91


roles/manage_agents/tasks/remove_agents.yml
```diff
@@ -0,0 +1,34 @@
...
+  ansible.builtin.set_fact:
+    managed_agents_removed_count: "{{ manage_agents_allocated.resources | length - manage_agents_desired_count | int }}"
+
+- name: Wait for all expected agents to be removed from the cluster
+  ansible.builtin.include_tasks:
+    file: allocated_agents.yml
+  with_sequence: start=1 end=10
+  loop_control:
+    pause: 30
+    break_when:
```
adriengentil, line 12: the pattern `until`/`retries`/`delay` doesn't work with `include_tasks`, but loop pattern look to be supported.

roles/manage_agents/tasks/add_agents.yml
```diff
@@ -0,0 +1,62 @@
...
+      kubernetes.core.k8s_json_patch:
+        api_version: agent-install.openshift.io/v1beta1
+        kind: Agent
+        name: "{{ item.metadata.name }}"
+        namespace: "{{ item.metadata.namespace }}"
+        patch:
+          - op: replace
+            path: /metadata/labels/{{ cluster_order_label | json_pointer_escape }}
+            value: "{{ manage_agents_cluster_order_name }}"
+      with_items: "{{ managed_agents_available.resources[: (managed_agents_selected_count | int)] }}"
```
larsks, line 50: I thought these days we prefer `loop` over `with_items` (primarily because `loop` doesn't have the magic nested-list-flattening behavior of `with_items`).

roles/manage_agents/tasks/add_agents.yml
```diff
@@ -0,0 +1,62 @@
...
+          {{ manage_agents_desired_count | int - manage_agents_allocated.resources | length }}
+
+    - name: Add selected agents to the cluster
+      kubernetes.core.k8s_json_patch:
+        api_version: agent-install.openshift.io/v1beta1
+        kind: Agent
+        name: "{{ item.metadata.name }}"
+        namespace: "{{ item.metadata.namespace }}"
+        patch:
+          - op: replace
```
larsks, line 47: Since we built `managed_agents_available` using a selector that excludes agents that already have `cluster_order_label`...

```
        label_selectors:
          - "!{{ cluster_order_label }}"
          - "{{ esi_agent_resource_class_label }}={{ manage_agents_resource_class }}"
```

...should this be `add` rather than `replace`? If you call `replace` and `path` doesn't exist, the patch is a no-op, and I think given the selector in the earlier task that the agents we're trying to label here will not have the `cluster_order_label` label.

plugins/filter/json_pointer_escape.py
larsks, line 1: For filter plugins that are not in a collection, Ansible (at least when running `ansible-playbook` from the command line) looks by default in the `filter_plugins` directory, rather than `plugins/filter`. Given a playbook `filtertest.yaml` like this in the top-level directory:

```
- hosts: localhost
  gather_facts: false
  tasks:
    - debug:
        msg: "{{ 'one/two/three' | json_pointer_escape }}"
```

Running `ansible-playbook filtertest.yaml` results in:

```
TASK [debug] *********************************************************************************************************************************
fatal: [localhost]: FAILED! => {"msg": "template error while templating string: Could not load \"json_pointer_escape\": 'json_pointer_escape'. String: {{ 'one/two/three' | json_pointer_escape }}. Could not load \"json_pointer_escape\": 'json_pointer_escape'"}
```

After running:

```
ln -s plugins/filter filter_plugins
```

The playbooks runs successfully:

```
TASK [debug] *********************************************************************************************************************************
ok: [localhost] => {
    "msg": "one~1two~1three"
}
```

I think we should either rename `filter/plugins` to `filter_plugins`, or -- preferably -- put the `json_pointer_escape` filter in a collection. I've created the `cloudkit.service` collection in #39 (merged) for things like this.

roles/manage_agents/tasks/add_agents.yml
```diff
@@ -0,0 +1,62 @@
...
+        namespace: "{{ default_agent_namespace }}"
+        label_selectors:
+          - "!{{ cluster_order_label }}"
+          - "{{ esi_agent_resource_class_label }}={{ manage_agents_resource_class }}"
+      register: managed_agents_available
+
+    - name: Count the number of agents to add to the cluster
+      ansible.builtin.set_fact:
+        managed_agents_selected_count: >
+          {{ manage_agents_desired_count | int - manage_agents_allocated.resources | length }}
```
larsks, line 38: There are some odd verb-tense issues going on here: we have some variables with `manage_` prefix and some with `managed_`. Given the role name, should we use `manage_` in all cases?

roles/manage_agents/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
...
+    namespace: "{{ default_agent_namespace }}"
+    label_selectors:
+      - "{{ cluster_order_label }}={{ manage_agents_cluster_order_name }}"
+      - "{{ esi_agent_resource_class_label }}={{ manage_agents_resource_class }}"
+  register: manage_agents_allocated
+
+- name: Remove agents from the cluster
+  ansible.builtin.include_tasks:
+    file: remove_agents.yml
+  when: (manage_agents_desired_count | int) < (manage_agents_allocated.resources | length)
```
larsks, line 15: Just curious: These casts to `int` are harmless, but are they necessary? Especially since we're running with `jinja2_native=True`.

plugins/filter/json_pointer_escape.py
adriengentil: Thanks for testing! There is so much magic when running `ansible-playbook` :face_exhaling:.

I tested my role without a playbook (with `ansible localhost --module-name include_role --args name=manage_agents`), and none the the paths worked, I had to set `ANSIBLE_FILTER_PLUGINS` manually.

roles/manage_agents/tasks/add_agents.yml
adriengentil: I tested a bit and `replace` worked (I understood it was just a delete+add operation).

But has you said, it makes more sense to use `add` in this flow since I check for the absence of the label before.

roles/manage_agents/tasks/add_agents.yml
adriengentil: makes sense, I try to follow `ansible-lint` recommendation and variables should be prefixed by  `manage_agents_`. I'll fix that.

roles/manage_agents/tasks/main.yml
adriengentil: Not sure we need to cast everywhere, but I remember I had some issues in the past, even with `jinja2_native`, where we sometimes still need to hint jinja that we want a `int` out of a variable.

roles/manage_agents/tasks/add_agents.yml
adriengentil: yes, the recommendation is to use `loop` https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html. I will update

adriengentil:
@larsks can you take a look again on this PR?

roles/manage_agents/tasks/add_agents.yml
```diff
@@ -0,0 +1,82 @@
...
+          {{ manage_agents_added.results | selectattr('changed') | list | length }}
+
+    - name: Print a message if not all agents were added
+      ansible.builtin.debug:
+        msg: "{{ manage_agents_added_count }} agents were added to the cluster instead of {{ manage_agents_selected_count }}. Please check agents availability."
+      when: manage_agents_added_count | int != manage_agents_selected_count | int
+
+  always:
+
+    - name: Delete agent lock
```
larsks, line 74: @mhrivnak pointed out in #38 that a failure here will leave stale leases floating around. If we include the ansible job id in the lease, we can the implement a cleanup job that looks for leases, checks if the job is active or not, and cleans them if necessary.

In #38 I have the job id encoded in the `holderIdentity`, but it probably makes more sense to add it as a label; maybe `cloudkit.openshift.io/aap-job-id`?

I'll make a new pr that applies that change to both places.

larsks (approved):

PR #38, opened 2025-05-06
Implement a locking mechanism to manage concurrency

larsks:
We would like to prevent multiple playbooks for the same cluster running
concurrently. This pull request implements a locking mechanism using
Kubernetes Lease objects [1]. If the create cluster playbook cannot acquire
a lock, it will exit immediately (and successfully).

This PR also revives the `cluster_settings` role in order to handle some
skew between the current template implementation and the fulfillment
service.

[1]: https://kubernetes.io/docs/concepts/architecture/leases/

### DIFF
--- a/playbook_cloudkit_create_hosted_cluster.yml
+++ b/playbook_cloudkit_create_hosted_cluster.yml
@@ -9,19 +9,93 @@
     cluster_order: "{{ ansible_eda.event.payload }}"
 
   pre_tasks:
+    - name: Set cluster order name
+      ansible.builtin.set_fact:
+        cluster_order_name: "{{ cluster_order.metadata.name }}"
+
+    - name: Set lock name and holder identity
+      ansible.builtin.set_fact:
+        cluster_order_lock_name: "{{ cluster_order_name }}-lock"
+        cluster_order_holder_id: >-
+          {{ cluster_order_name }}-{{ awx_job_id | default("unknown") }}-{{
+            lookup('community.general.random_string', special=false, numbers=false, upper=false)
+          }}
+
+    - name: Apply default settings
+      ansible.builtin.include_role:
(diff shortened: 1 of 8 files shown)


tzumainn (commented):
Is it possible to rename `cluster_order_lease_name` to `cluster_order_name`, or is the `lease_name` suffix by kubernetes leases?

larsks:
> Is it possible to rename `cluster_order_lease_name` to `cluster_order_name`, or is the `lease_name` suffix by kubernetes leases?

I'm not sure I understand the question; there is already a `cluster_order_name` (which is the name of the clusterorder manifest), and it is different from `cluster_order_lease_name` (which is the name of the lease).

larsks:
> Okay - in that case I'd argue that we'd want to rename the task names as well

Makes sense; I've made those changes.

tzumainn (approved):
Looks good - thanks!

adriengentil (approved):

playbook_cloudkit_create_hosted_cluster.yml
```diff
@@ -9,17 +9,93 @@
...
+
+        # Without this, a failure in the template would show up as
+        # a successful job run.
+        - name: Propogate failure
+          ansible.builtin.fail:
+            msg: Propogating earlier failure from rescue block
+
+      always:
+
+        - name: Delete lock
```
mhrivnak, line 92: What happens if execution fails right here? Will anything clean up the lock?

larsks: Nothing will right now. Since the lase includes the AAP job id, we could write a lease-cleaning job that finds leases, checks to see if the job is still active, and deletes them if not.

I've opened innabox/issues#131

PR #40, opened 2025-05-09
Destroy ingress floating ip

larsks:
When deleting a cluster, we were previously only deleting the api floating
ip. With this commit we will also clean up the ingress floating ip.


tzumainn (approved):
Ah, yes, makes sense!

PR #44, opened 2025-05-12
Add role to publish templates to fulfillment service

larsks:
The `cloudkit.service.publish_templates` role will publish the templates in
`cloudkit_templates` to the fulfillment service. This commit includes
changes to the `find_template_roles` filter and to the `cloudkit.yaml`
metadata files so that they match the requirements of the fulfillment
service.

### DIFF
--- a/collections/ansible_collections/cloudkit/service/roles/publish_templates/tasks/main.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/publish_templates/tasks/main.yaml
@@ -1,0 +1,41 @@
+- name: Get list of existing templates
+  ansible.builtin.uri:
+    url: "{{ publish_templates_api_endpoint }}"
+    headers:
+      authorization: Bearer {{ cloudkit_fulfillment_service_token }}
+  register: template_check
+
+- name: Create list of existing template template ids
+  ansible.builtin.set_fact:
+    template_ids: >-
+      {{
+      template_check|json_query("json.items[].id")
+      }}
+
+- name: Update existing template
+  loop: "{{ cloudkit_templates }}"
+  loop_control:
+    label: "{{ template.id }}"
+    loop_var: template
+  when: template.id in template_ids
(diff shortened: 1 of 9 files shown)


larsks:
@jhernand I'm running into an error because I'm a little unsure of how to serialize the `default` field for a template parameter. In the protocol it uses the `protobufAny` schema; would this be appropriate for a string default value?

```json
{
  "default": {
    "@type": "type.googleapis.com/google.protobuf.StringValue",
	"value": "some_string_Value"
  }
}
```

larsks:
It looks like that was correct; the publish role no longer errors out on parameters with default values. We only support scalar defaults for now, since that's a relatively simple mapping from `type(variable)` to the appropriate protobuf string.

jhernand (approved):

collections/ansible_collections/cloudkit/service/roles/publish_templates/tasks/main.yaml
```diff
@@ -0,0 +1,43 @@
+- name: Check for existing templates
+  loop: "{{ cloudkit_templates }}"
+  loop_control:
+    label: "{{ item.id }}"
+  ansible.builtin.uri:
+    url: "{{ publish_templates_api_endpoint }}/{{ item.id }}"
```
adriengentil, line 6: don't we have an endpoint to list the templates instead of getting them one by one? It would also help for the deletion/deprecation. Anyway, I don't think it's a blocker for the moment.

adriengentil (approved):

collections/ansible_collections/cloudkit/service/roles/publish_templates/tasks/main.yaml
larsks: > don't we have an endpoint to list the templates instead of getting them one by one?

We do, and I've made that change to the playbook. Take a look and let me know if that seems reasonable.

PR #50, opened 2025-05-19
Require python-esiclient >= 1.3.0

larsks:
We use the `openstack esi port forwarding` command in the `external_access`
role to configure port forwarding. Previously, this command would fail with
an error when attempting to create a port forwarding that already exists.
This made it difficult to identify legitimate failures.

With cci-moc/python-esiclient#89, python-esiclient is now idempotent and
will not emit an error if an existing port forwarding exactly matches the
requested configuration.


tzumainn (approved):
Can confirm that's what it does!

PR #54, opened 2025-05-20
Prevent Ansible from number-izing our labels

larsks:
Ansible is unhelpfully converting a string into a number. We're asking for
this:

    labels:
      cloudkit.openshift.io/aap-job-id: "{{ awx_job_id|default('unknown') }}"

And Ansible is producing something like this:

    labels:
      cloudkit.openshift.io/aap-job-id: 1234

Which fails because label values can only be strings. We can prevent this
behavior by ensuring that the label value cannot be interpreted as a
number:

    labels:
      cloudkit.openshift.io/aap-job-id: "job-{{ awx_job_id|default('unknown') }}"

### DIFF
--- a/roles/manage_agents/tasks/add_agents.yml
+++ b/roles/manage_agents/tasks/add_agents.yml
@@ -16,7 +16,7 @@
             name: "agent-{{ manage_agents_resource_class }}-lock"
             namespace: "{{ default_agent_namespace }}"
             labels:
-              cloudkit.openshift.io/aap-job-id: "{{ awx_job_id|default('unknown') }}"
+              cloudkit.openshift.io/aap-job-id: "job-{{ awx_job_id | default('unknown') }}"
           spec:
             holderIdentity: "{{ cluster_order_holder_id }}"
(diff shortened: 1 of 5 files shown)


roles/manage_agents/tasks/add_agents.yml
```diff
@@ -16,7 +16,7 @@
             name: "agent-{{ manage_agents_resource_class }}-lock"
             namespace: "{{ default_agent_namespace }}"
             labels:
-              cloudkit.openshift.io/aap-job-id: "{{ awx_job_id|default('unknown') }}"
+              cloudkit.openshift.io/aap-job-id: "job-{{ awx_job_id | default('unknown') }}"
```
adriengentil, line 19:
```suggestion
              cloudkit.openshift.io/aap-job-id: "{{ awx_job_id | default('unknown') | str }}"
```

have you tried cast to `str`, I had to do the opposite in the Nodepool to force an `int` instead to a `str`

adriengentil (approved):
just one comment, but it's not blocking

PR #57, opened 2025-05-21
Configure agents when reconciling hosted cluster

adriengentil:


roles/hosted_cluster/tasks/nodes.yml
```diff
@@ -0,0 +1,57 @@
...
+    kubernetes.core.k8s_json_patch:
+      api_version: agent-install.openshift.io/v1beta1
+      kind: Agent
+      name: "{{ item.metadata.name }}"
+      namespace: "{{ item.metadata.namespace }}"
+      patch:
+        - op: add
+          path: /spec/approved
+          value: true
+    loop: "{{ manage_agents_added_result }}"
```
adriengentil, line 51: we should approve all agents with the label in order to be idempotent

adriengentil: the approval can be done once in the `main.yml`

roles/hosted_cluster/tasks/main.yml
```diff
@@ -78,26 +78,82 @@
...
+  loop_control:
+    label: "Manage NodePool nodepool-{{ hosted_cluster_name }}-{{ item.resource_class }}"
+
+- name: Create cluster network
+  ansible.builtin.include_role:
+    name: massopencloud.esi.network
+  vars:
+    network_suffix: "{{ hosted_cluster_name }}"
+    network_state: "{{ hosted_cluster_state }}"
+  when: hosted_cluster_state == "present"
```
larsks, line 115: Why does only this one task of a `when` condition on `hosted_cluster_state`?

roles/hosted_cluster/tasks/main.yml
```diff
@@ -78,26 +78,82 @@
...
+
+- name: Wait for the Agents to be removed from the cluster
+  ansible.builtin.include_role:
+    name: manage_agents
+    tasks_from: wait_for_agents_to_be_removed
+  vars:
+    manage_agents_cluster_order_name: "{{ hosted_cluster_name }}"
+    manage_agents_desired_count: "{{ item.number_of_nodes | int }}"
+    manage_agents_resource_class: "{{ item.resource_class }}"
+  loop: "{{ hosted_cluster_node_requests }}"
```
larsks, line 143: I am a little confused about what is happening here...if we are creating a clsuter, why are we waiting for agents to be removed?

larsks (commented):
I have a couple of questions, and there are a couple of ansible-lint warnings that look like they could be legitimate.

roles/hosted_cluster/tasks/main.yml
adriengentil: the `hosted_cluster` manages creation and deletion of clusters based on `hosted_cluster_state`

roles/hosted_cluster/tasks/main.yml
adriengentil: if create (`hosted_cluster_state == "present"`) a hosted cluster, we want to create a network before any agent are attached to it, if we delete (`hosted_cluster_state == "absent"`) a hosted cluster I guess we want to detach all the agents from it before we delete it (the deletion task is at the end of the file).

roles/manage_agents/tasks/detach_and_unlabel_all_removed_agents.yml
```diff
@@ -1,31 +1,29 @@
...
+- name: Detach removed agents from cluster network
+  ansible.builtin.include_role:
+    name: massopencloud.esi.detach_network_from_node
+  vars:
+    node: "{{ item.metadata.annotations['esi.nerc.mghpcc.org/uuid'] }}"
+    network: "{{ manage_agents_cluster_network }}"
+  loop: "{{ manage_agents_removed }}"
+  loop_control:
+    label: "Detach agent {{ item.metadata.name }}"
+
```
tzumainn, line 25: Just a note that we probably want to attach these agents back to the default `idle_agents` network, although it's unclear to me whether that network should be a default or not.

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small_github/meta/cloudkit.yaml
```diff
@@ -7,12 +7,12 @@ description: >
...
-  number_of_nodes: 2
+- resourceClass: fc430
+  numberOfNodes: 2
 
 # This declares that for a cluster built from this template, only the resource
 # classes listed here may be used when requesting additional nodes. If this
 # list is empty or unspecified, a ClusterOrder may request nodes from any
 # available resource class.
-allowed_resource_classes:
+allowed_resourceClasses:
```
tzumainn, line 17: super minor note that the casing of this attribute feels weird, with its combination of underscores and camel casing; maybe worth changing now since I don't think this is actually used anywhere?

roles/external_access/tasks/create_external_access.yaml
```diff
@@ -66,17 +66,32 @@
...
+- name: Get esi node name of an agent
+  ansible.builtin.set_fact:
+    external_access_worker_node_name: >-
+      {{
+      agents.resources[0].spec.hostname
+      }}
+
+- name: Get network details from esi
+  ansible.builtin.command: >-
+    openstack esi node network show {{ external_access_worker_node_name }} -f json
```
tzumainn, line 78: Not something that needs to be fixed (especially since I think this is all eventually going to be refactored to use a bare metal service), but I think `openstack esi node network list --node MOC-R4PAC24U35-S3A --network network-larstest2` might have gotten what you wanted!

roles/manage_agents/tasks/detach_and_unlabel_all_removed_agents.yml
```diff
@@ -1,31 +1,29 @@
...
 
 - name: List agents removed from the cluster
   ansible.builtin.set_fact:
     manage_agents_removed: >
       {{ manage_agents_allocated.resources | selectattr('spec.clusterDeploymentName.name', 'undefined') }}
 
-- name: Cleanup removed agents
+- name: Detach removed agents from cluster network
+  ansible.builtin.include_role:
+    name: massopencloud.esi.detach_network_from_node
```
tzumainn, line 18: Do we want to update to use the new network/node playbook?

tzumainn (commented):
Some brief comments!

roles/manage_agents/tasks/detach_and_unlabel_all_removed_agents.yml
larsks: Yeah, probably, hadn't actually tested the delete code yet! Let me take a look.

roles/external_access/tasks/create_external_access.yaml
larsks: I've made this change; I guess it simplifies the expression required to extract the node ip.

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small_github/meta/cloudkit.yaml
larsks: Too much `xargs sed ...`.

tzumainn (approved):
It's all so beautiful

PR #64, opened 2025-05-30
Resolve publish job failures

larsks:
Due to recent changes in our playbooks, we need to set validation aliases
on the fields in the NodeRequest class.

And while we're at it, we set a sane default value for
`cloudkit_template_collections`.


tzumainn (approved):
Ah! Gotcha

PR #70, opened 2025-05-31
Refactor infrastructure management roles

larsks:
This commit reorganizes some of the infrastructure setup/teardown so that
(a) we can properly sequence setup/teardown of the port forwarding and dns
records with cluster teardown, and (b) to resolve some errors caused by
sharing the same workflows for both create and delete operations.

In particular, this moves the agent and network setup out of the
`hosted_cluster` role and into the `cluster_infra` role.


tzumainn (approved):
I like this change - for me at least, it makes it clear what's happening when a cluster is being created or destroyed.

PR #74, opened 2025-06-10
Use license manifest instead of RH credentials

adriengentil:
We now need to use service accounts in console.redhat.com to fetch a
license, but it did not work for me. We can just upload a license
manifest as an alternative method.

Required to move forward with https://github.com/innabox/issues/issues/135


collections/ansible_collections/cloudkit/config_as_code/README.md
```diff
@@ -201,21 +209,24 @@ oc apply -f cloudkit_env.yml -n fulfillment-aap
...
+aap_username: ...
+aap_password: ...
+aap_organization_name: ...
+aap_project_name: ...
+aap_project_git_uri: ...
+aap_project_git_branch: ...
+aap_ee_image: ...
+aap_validate_certs: ...
+license_manifest_path: ...
+```
```
larsks, line 225: This seems to substantially duplicate the information on line 183. Do we need to have this in two separate places?

collections/ansible_collections/cloudkit/config_as_code/README.md
```diff
@@ -257,6 +269,14 @@ spec:
                 secretKeyRef:
                   name: fulfillment-admin-password
                   key: password
+            - name: CONFIG_AS_CODE_EXTRA_VARS_PATH
+              value: /var/secrets/config-as-code/vars.yaml
```
larsks, line 273: This section should be updated to document the format of the secret. This value implies that the secret has a single key, `vars.yaml`, that contains the content from "option 1", but you're never explicit on that point. Showing how to create the secret from the file created earlier might be a good idea (e.g., showing the `oc create secret generic config-as-code --from-file=...` command line).

collections/ansible_collections/cloudkit/config_as_code/roles/load_extra_vars/tasks/main.yml
```diff
@@ -0,0 +1,21 @@
...
+
+- name: Render extra vars template
+  ansible.builtin.template:
+    src: "{{ load_extra_vars_file_path }}"
+    dest: "{{ extra_vars.path }}"
+    mode: "0644"
+
+- name: Load extra vars
+  ansible.builtin.include_vars:
+    file: "{{ extra_vars.path }}"
```
larsks, line 16: Why do we do this instead of simply:

```
- name: Load extra vars
  ansible.builtin.include_vars:
    file: "{{ load_extra_vars_file_path }}"
```

?

adriengentil: `ansible.builtin.include_vars` alone does not render jinja templates (e.g. : the `{{ lookup('env', 'AAP_PASSWORD') }}`).

adriengentil:
@jkary I changed this PR to rely on 2 secrets, I did not test yet, but can you have a first look in the meantime?

jkary (approved):

PR #78, opened 2025-06-16
add ai-cluster-ansible collection to requirements.yml 

IsaiahStapleton:
https://github.com/innabox/issues/issues/155

### DIFF
--- a/collections/requirements.yml
+++ b/collections/requirements.yml
@@ -12,3 +12,7 @@ collections:
   - name: "cloudkit.config_as_code"
     source: "collections/ansible_collections/cloudkit/config_as_code"
     type: dir
+  - name: "isaiahstapleton.ai_cluster_ansible"
+    source: "https://github.com/IsaiahStapleton/ai-cluster-ansible.git"
+    type: git
+    version: main


larsks (approved):
This looks fine. I'm building it locally to confirm everything works; I'll merge this as soon as the build completes.

adriengentil:
@IsaiahStapleton would it be possible to tag your repo with a version and use that version in `requirements.yml`?

Using `main` as a version causes 2 issues:

- the installation of the collection works the first time, but fails to re-install if the collection is already in Ansible's collection path: https://github.com/ansible/ansible/issues/81687, I tried to fix with https://github.com/innabox/cloudkit-aap/pull/79/commits/166c9efd456902e330682e233c07d84186335e9e
- each time `ansible-galaxy collection install ...` is called, it will pull the git repo to get the last version, even though the collection is already there in the collection path. It's a problem in the scope of my proposal in PR https://github.com/innabox/cloudkit-aap/pull/79 where I vendor the dependencies in our repo. It means that the vendored collection in the repo may not be the one included in the execution environment (the container image used by AAP jobs), which is an issue as one of the goal is to make the build of the execution environment reproducible. Unfortunately, I can't disable the galaxy step in `ansible-builder` without an ugly hack

PR #79, opened 2025-06-17
Add the automation to build the execution environment

adriengentil:
This PR make the execution environment easier to build, and it doesn't require any Red Hat credentials to build:
- Use centos streams instead of the image behind registry.redhat.io
- Remove the dependency on `openshift-client` RPM, use the binaries contained in the public OCI image of OCP instead
- Vendor the collections we depend on (they are either GPL or Apache, so it should be fine to re-distribute)
- Create  a GHA workflow:
  - Build on PR
  - Build and push to GHCR on push

https://github.com/innabox/cloudkit-aap/issues/46


adriengentil:
> I understand why we're vendoring the redhat collections, but what are the advantages of vendoring everything else?

I did it mainly because the tooling works like that, otherwise we would need 2 requirement files.

At least the vendoring acts like a lock, if some dependencies are loosely defined, then what is vendored is what will end-up being executed in AAP jobs.

PR #80, opened 2025-06-23
Use tag instead of branch for ansible collection

IsaiahStapleton:
As per the request here: https://github.com/innabox/cloudkit-aap/pull/78#issuecomment-2983410621


adriengentil (approved):
Thanks!

PR #81, opened 2025-06-26
Rename arguments in hosted_cluster argument spec

larsks:
roles/hosted_cluster/meta/argument_specs.yaml had some incorrect
parameter names, causing erroneous playbook failures.


tzumainn:
Do these parameter names have to match those in https://github.com/innabox/cloudkit-aap/blob/main/collections/ansible_collections/cloudkit/service/plugins/filter/find_template_roles.py#L151?

PR #84, opened 2025-07-10
add workflow to perform post-install tasks

adriengentil:


collections/ansible_collections/cloudkit/service/plugins/filter/to_temp_file.py
```diff
@@ -0,0 +1,37 @@
...
+        content: The content to write to the temporary file
+        suffix: Optional suffix for the temporary file (e.g., '.yaml', '.json')
+        prefix: Optional prefix for the temporary file
+        dir: Optional directory to create the temp file in
+    
+    Returns:
+        str: The path to the created temporary file
+    
+    Example:
+        {{ kubeconfig_content | temp_file(suffix='.yaml') }}
```
larsks, line 22: This example should read `... | to_temp_file`.

playbook_cloudkit_create_hosted_cluster_post_install.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Cloudkit hosted cluster post-install configuration
+  hosts: localhost
+  connection: local
+  gather_facts: false
+
+  environment:
+    KUBECONFIG: "{{ kubeconfig | cloudkit.service.to_temp_file }}"
```
larsks, line 8: I like this solution.

playbook_cloudkit_create_hosted_cluster_post_install.yml
```diff
@@ -0,0 +1,22 @@
...
+  pre_tasks:
+    - name: Extract template info
+      ansible.builtin.include_role:
+        name: extract_template_info
+
+  tasks:
+    - name: Post-install hosted cluster configuration
+      ansible.builtin.include_role:
+        name: "{{ template_id }}"
+        tasks_from: post_install.yaml
```
larsks, line 22: For consistency, hould we update `playbook_cloudkit_create_hosted_cluster.yml` to explicitly use `install.yaml` rather than defaulting to `main.yaml`?

larsks: On second thought, maybe not, since it performs both install and teardown activities based on parameters.

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small/tasks/main.yaml
```diff
@@ -1,5 +1,5 @@
 - name: Create a new instance of this template
-  when: cluster_template_state == 'present'
+  when: cluster_template_state == 'present' and cluster_template_stage == 'install'
```
larsks, line 2: We're not setting `cluster_template_stage` anywhere, so I think this conditional will mean we never run the installation tasks.

playbook_cloudkit_create_hosted_cluster_post_install.yml
adriengentil: For the sake of consistency, I was thinking about removing `state` parameter, and instead include `install.yml`, `post-install.yml` or `delete.yml` depending on the context.

collections/ansible_collections/cloudkit/templates/roles/ocp_4_17_small/tasks/install.yaml
```diff
@@ -28,3 +28,10 @@
     external_access_state: present
     external_access_namespace: "{{ cluster_working_namespace }}"
     external_access_ingress_internal_network: "network-{{ cluster_order.metadata.name }}"
+
+- name: Retrieve kubeconfig
+  ansible.builtin.include_role:
+    name: retrieve_kubeconfig
+  vars:
+    retrieve_kubeconfig_cluster_name: "{{ hosted_cluster_name }}"
+    retrieve_kubeconfig_namespace: "{{ cluster_working_namespace }}"
```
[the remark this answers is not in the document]
tzumainn: ah, got it!

PR #94, opened 2025-08-08
Remove hardcoded namespace and service account

larsks:
The publish templates playbook had a hardcoded namespace name and service
account name. This commit replaces those with variables that get default
values from the `CLOUDKIT_PUBLISH_TEMPLATES_NAMESPACE` and
`CLOUDKIT_PUBLISH_TEMPLATES_SERVICE_ACCOUNT` environment variables.

The config-as-code roles are modified to set these environment variables
using `fieldRef` expressions [1] on the pod that runs the publish templates
job.

[1]: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/


larsks:
Never mind; converted to draft because `spec.serviceAccountName` is in fact the wrong service account to use for authenticating against the fulfillment service.

PR #109, opened 2025-08-17
Fix cluster create workflow

tzumainn:
The syntax used for specifying dependent steps was out-of-date; this PR updates that with `related`.


tzumainn:
This is where I found the information: https://github.com/redhat-cop/infra.aap_configuration/issues/304

larsks (approved):

PR #121, opened 2025-09-02
Feature/ocp virt vm template

jkary:
Initial VMaaS working with basic VM definition.


adriengentil:
I think it might be good to isolate VM-related roles to their own collection. Same for the templates.

collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/meta/cloudkit.yaml
```diff
@@ -0,0 +1,21 @@
+# Define the display name and description of the template
+title: OpenShift Virtualization VM
```
adriengentil, line 2: nit: I'm not sure we want user to know it's running on OCP. We should just say that's a VM.

collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/defaults/main.yaml
```diff
@@ -0,0 +1,128 @@
+---
+# Default VM configuration values
+
+# VM Resource Defaults
+default_vm_cpu_cores: 2
+default_vm_memory: "2Gi"
```
adriengentil, line 6: should we let users to specify cpu+memory, or to reference instance classes?

collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/tasks/install.yaml
```diff
@@ -0,0 +1,217 @@
...
+
+- name: Display VM information
+  ansible.builtin.debug:
+    msg:
+      - "Virtual Machine '{{ vm_name }}' created successfully"
+      - "Namespace: {{ vm_namespace }}"
+      - "CPU Cores: {{ vm_cpu_cores }}"
+      - "Memory: {{ vm_memory }}"
+      - "Root Disk Size: {{ vm_disk_size }}"
+      - "Status: {{ vm_status.resources[0].status.printableStatus | default('Unknown') }}"
```
adriengentil, line 217: should we create a service to expose SSH at least, and be able to log on it?

collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/meta/cloudkit.yaml
```diff
@@ -0,0 +1,21 @@
...
+description: >
+  This template provisions a virtual machine using OpenShift Virtualization 
+  (KubeVirt) on an existing OpenShift cluster. It creates the VM with 
+  configurable CPU, memory, and storage resources.
+
+# This sets the nodeRequest field in the ClusterOrder manifest. For VM templates,
+# this represents the underlying OpenShift cluster requirements rather than 
+# bare metal nodes.
+default_node_request:
+- resourceClass: ocp_cluster
```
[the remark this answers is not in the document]
jkary: This should be removed.  It was there when I was originally adding VMaaS as part of the cluster workflow.  It does occur to me that the concept of having an ocp_cluster could be very useful for specific deployment cluster targeting.

PR #123, opened 2025-09-04
Manage infrastructure finalizer on clusterorder

larsks:
Add and remove a finalizer on cluster orders so that the cluster order is
not deleted until the associated infrastructure has been deleted. This
allows us to recover from failures and run the delete workflow more than
once. Previously, the delete workflow could only run once even if it failed
because the cluster order would get deleted.

This partially address innabox/issues#208

### DIFF
--- a/playbook_cloudkit_create_hosted_cluster.yml
+++ b/playbook_cloudkit_create_hosted_cluster.yml
@@ -44,7 +44,6 @@
   tasks:
     - name: Create cluster
       block:
-
         # By using service_side_apply and registering this job run as the
         # field manager, we ensure that future attempts to create/modify
         # this lock will fail (because of the conflict setting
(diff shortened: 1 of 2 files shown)


collections/ansible_collections/cloudkit/service/roles/manage_finalizer/tasks/main.yaml
```diff
@@ -0,0 +1,19 @@
+- name: Get current ClusterOrder state
+  ansible.builtin.set_fact:
+    manage_finalizer_cluster_order: >-
+      {{
+        lookup('kubernetes.core.k8s', api_version=cluster_order.apiVersion, kind=cluster_order.kind,
+          resource_name=cluster_order.metadata.name, namespace=cluster_order.metadata.namespace)
+      }}
+
+- name: Add finalizer to ClusterOrder
+  when: manage_finalizer_state|default('') == "present"
```
tzumainn, line 10: random question; is the default considered a good thing to have, even if the parameter is marked as `required` in the argument_specs?

jkary (approved):
Just need to resolve the GitHub actions issues.

PR #126, opened 2025-09-09
Log configuration in publish templates playbook

larsks:
Showing current configuration values can aid in problem diagnosis.


tzumainn (approved):
Looks good!

PR #133, opened 2025-09-30
Add ClusterTemplate node sets to python filter

tzumainn:
This change converts the 'default_node_requests' attribute in Ansible playbook templates into ClusterTemplate 'node_sets'

### DIFF
--- a/collections/ansible_collections/cloudkit/service/plugins/filter/find_template_roles.py
+++ b/collections/ansible_collections/cloudkit/service/plugins/filter/find_template_roles.py
@@ -155,6 +155,13 @@ class NodeRequest(Base):
     number_of_nodes: int = pydantic.Field(..., validation_alias="numberOfNodes")
 
 
+class NodeSet(Base):
+    """NodeSet represents the template's default bare metal resources"""
+
+    host_class: str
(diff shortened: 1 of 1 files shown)


tzumainn:
Tested by taking the generated JSON and successfully creating a ClusterTemplate; then I created a Cluster using the template and verified it had the following in its CRD:

```
spec:
  nodeRequests:
    - numberOfNodes: 2
      resourceClass: fc430
    - numberOfNodes: 1
      resourceClass: fc830
  templateID: mainn
  templateParameters: '{}'
```

collections/ansible_collections/cloudkit/service/plugins/filter/find_template_roles.py
```diff
@@ -188,6 +188,12 @@ def serialize_path(self, value: Path):
     def id(self) -> str:
         return f"{self.collection}.{self.name}"
 
+    @pydantic.computed_field
+    def node_sets(self) -> dict:
```
adriengentil, line 192:
```suggestion
    def node_sets(self) -> dict[str, dict[str, Any]]:
```
to fully defined the return type

PR #137, opened 2025-10-15
Fix cluster scale down when resource class is completely removed

tzumainn:
The cluster scaledown playbook did not work when a resource class was completely removed, as it only iterated over the resource classes in the incoming request. This PR fixes that by doing the following:

* updating NodePool creation to include a resource class label
* changing the scaledown playbook to
  * compare existing NodePool resource classes against the resource classes in the request
  * updating the node request array to include a 0-count entry for the unneeded resource classes

### DIFF
--- a/collections/ansible_collections/cloudkit/service/roles/cluster_infra/tasks/create_cluster_infra.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/cluster_infra/tasks/create_cluster_infra.yaml
@@ -19,6 +19,37 @@
 # we run detach_and_unlabel_all_removed_agents, causing them to be
 # erroneously detached.
 
+- name: Retrieve all currently allocated node pools
+  kubernetes.core.k8s_info:
+    api_version: hypershift.openshift.io/v1beta1
+    kind: NodePool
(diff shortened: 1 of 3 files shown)


tzumainn:
Marking as draft for now since I haven't yet tested this on a live cluster, although I have tested the basic theory against random resources in the dev environment. Right now, I'm mostly hoping that someone can tell me if there's an easier way to do some of these Ansible tasks I updated

tzumainn:
> @adriengentil @larsks I think this is ready for review. I tested by editing a ClusterOrder CRD directly; there are other issues that prevent doing this through the fulfillment-service, but I figure that's separate. The cluster order also doesn't actually scale down - there's a wait task in https://github.com/innabox/cloudkit-aap/blob/main/collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/wait_for_agents_to_be_removed.yml that loops and the fails - but I think that's also a separate issue.

Nevermind - adrien helped me understand that there's a task that manages the NodePool, and there's a task that waits for the infrastructure to be ready. I'll have to update the task that manages the NodePool to remove unwanted NodePools.

PR #139, opened 2025-10-18
Check if agent label exists before running k8s_json_patch remove oper…

tzumainn:
The k8s_json_patch remove operation will fail if the specified path doesn't exist. This PR changes the remove tasks to first filter agents by whether the path exists before running the remove operation.


tzumainn:
I couldn't figure out a simple way to filter by labels that use dot notation; let me know if there's something I missed!

adriengentil (approved):
just a suggestion, but it's ok merge like that :+1:

Issue #130, opened 2025-09-26
Add ownerReference to Lease


larsks:
To ensure that leases get cleaned up properly, add an ownerReference to the pod running the Ansible job. This should cause kubernetes to garbage collect the lease when the job exits.

This resolves the situation in which a problem with the ansible job prevents the playbook from cleaning up the lease.

PR #142, opened 2025-10-22
Use ownerReference to manage lease

larsks:
Relying on a block/rescue/always construct to manage the lease was problematic for
at least two reasons:

- If a job was cancelled, this could prevent the lease cleanup code from
  running.
- Some Ansible errors are un-catchable, and cause Ansible to abort
  immediately rather than executing the `rescue` or `always` blocks [1]

In either case, the job would exit without cleaning up the lease, which
would in turn prevent any future job for the same target cluster from
successfully executing.

We resolve this by replacing the block/rescue/always construct with an
`ownerReference` [2] on the lease. This causes Kubernetes to remove the
lease for us automatically when the pod exits.

Because ownerReferences cannot cross namespaces, this requires creating the
Lease in the same namespaces as the AAP job, rather than in the "cluster
working namespace".

[1]: https://github.com/innabox/cloudkit-aap/issues/127
[2]: https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

Closes: #130

### DIFF
--- a/collections/ansible_collections/cloudkit/service/roles/lease/meta/argument_specs.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/lease/meta/argument_specs.yaml
@@ -1,0 +1,11 @@
+argument_specs:
+  main:
+    options:
+      lease_name:
+        type: str
+        required: true
+      lease_holder:
+        type: str
+        required: true
+      lease_labels:
+        type: dict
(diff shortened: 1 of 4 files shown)


larsks:
> we also have a lease here that might need GC if something does wrong: 

I updated the agent lease.

PR #147, opened 2025-10-23
Add finalizer to clusterorder for infra cleanup

larsks:
We don't start cleaning up the ESI infrastructure until after the
HostedCluster has been destroyed. This is also what triggers the operator
to delete the ClusterOrder resource. Previously, this meant that if there
was a failure cleaning up ESI resources, we would never have a chance to
retry the delete: with the ClusterOrder deleted, the delete webhook will
never receive another notification for the cluster.

By adding a finalizer to the ClusterOrder, we ensure that it hangs around
until we have successfully completely infrastructure cleanup.

Part of: https://github.com/innabox/issues/issues/239


tzumainn (approved):
Makes sense; also successfully tested.

PR #155, opened 2025-10-30
Add playbooks for adding/removing ESI nodes as agents

ajamias:
There is currently not an automated way to add/remove ESI nodes as agents to the hostpool. By adding these playbooks and the `massopencloud.esi.agent` role, a user can specify which ESI nodes they want to include and exclude in the hostpool.

Part of: https://github.com/innabox/issues/issues/195

### DIFF
--- a/collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
@@ -1,0 +1,131 @@
+- name: Display node names that will be added as agents
+  ansible.builtin.debug:
+    var: manage_agents_node_names
+
+- name: Get all agents
+  kubernetes.core.k8s_info:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+  register: initial_agents_result
+
+- name: Extract initial agents
+  ansible.builtin.set_fact:
+    initial_agents: "{{ initial_agents_result.resources }}"
+
+- name: Filter non-agent node names
+  when: initial_agents | length > 0
+  ansible.builtin.set_fact:
+    manage_agents_node_names: >-
+      {{
+        manage_agents_node_names |
+        difference(initial_agents |
+          selectattr('spec.hostname', 'defined') |
+          map(attribute='spec.hostname') |
+          list
+        )
+      }}
+
+- name: Get combined node and port info
+  ansible.builtin.include_role:
+    name: massopencloud.esi.node
+    tasks_from: get_nodes
+  vars:
+    node_filter_names: "{{ manage_agents_node_names }}"
+# set_fact: node_info_list
+
+- name: Get InfraEnv information
+  kubernetes.core.k8s_info:
+    kind: InfraEnv
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+    name: "{{ manage_agents_infraenv_name }}"
+  register: infraenv
+
+- name: Set discovery_url
+  ansible.builtin.set_fact:
+    discovery_url: "{{ infraenv.resources[0].status.isoDownloadURL }}"
+  when: not discovery_url|default(false)
+
+- name: Provision nodes
+  when: >-
+    node_info.ports |
+    map(attribute='address') |
+    cloudkit.service.mac_to_agent_name(initial_agents) is none
+  ansible.builtin.include_role:
+    name: massopencloud.esi.node
+    tasks_from: provision_node
+  vars:
+    node_name: "{{ node_info.name }}"
+    node_provisioning_network_name: "{{ manage_agents_provisioning_network_name }}"
+    node_discovery_url: "{{ discovery_url }}"
+    node_agent_namespace: "{{ manage_agents_namespace }}"
+    node_deploy_wait: false
+  loop: "{{ node_info_list }}"
+  loop_control:
+    loop_var: node_info
+    label: "{{ node_info.name }}"
+
+- name: Wait for nodes to register as agents
+  kubernetes.core.k8s_info:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+  register: current_agents
+  until: >
+    node_info.ports |
+    map(attribute='address') |
+    cloudkit.service.mac_to_agent_name(current_agents.resources) is not none
+  retries: 90
+  delay: 10
+  loop: "{{ node_info_list }}"
+  loop_control:
+    loop_var: node_info
+    label: "{{ node_info.name }}"
+
+- name: Get all agents
+  kubernetes.core.k8s_info:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+  register: final_agents_result
+
+- name: Extract final agents
+  ansible.builtin.set_fact:
+    final_agents: "{{ final_agents_result.resources }}"
+
+- name: Configure agent metadata for all nodes
+  kubernetes.core.k8s:
+    state: present
+    definition:
+      apiVersion: agent-install.openshift.io/v1beta1
+      kind: Agent
+      metadata:
+        name: "{{ agent_metadata.name }}"
+        namespace: "{{ manage_agents_namespace }}"
+        annotations: "{{ agent_metadata.annotations }}"
+        labels: "{{ agent_metadata.labels }}"
+      spec:
+        approved: true
+        hostname: "{{ agent_metadata.hostname | lower }}"
+  loop: >-
+    {{
+      node_info_list |
+      massopencloud.esi.get_agent_metadata(final_agents)
+    }}
+  loop_control:
+    loop_var: agent_metadata
+    label: "{{ agent_metadata.name }}"
+
+- name: Move agents to idle agents network
+  ansible.builtin.include_role:
+    name: massopencloud.esi.l2
+    tasks_from: set_networks_for_node
+  vars:
+    node: "{{ node_info.name }}"
+    networks:
+    - "{{ manage_agents_idle_agents_network }}"
+  loop: "{{ node_info_list }}"
+  loop_control:
+    loop_var: node_info
+    label: "{{ node_info.name }} -> {{ manage_agents_idle_agents_network }}"
--- a/collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/remove_agents.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/remove_agents.yaml
@@ -1,0 +1,51 @@
+- name: Display node names that will be removed from being agents
+  ansible.builtin.debug:
+    var: manage_agents_node_names
+
+- name: Get agents
+  kubernetes.core.k8s_info:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+  register: agents_result
+
+- name: Extract agents
+  ansible.builtin.set_fact:
+    initial_agents: "{{ agents_result.resources }}"
+
+- name: Get node list for removal
+  ansible.builtin.include_role:
+    name: massopencloud.esi.node
+    tasks_from: get_nodes
+  vars:
+    node_filter_names: "{{ manage_agents_node_names }}"
+# set_fact: node_info_list
+
+- name: Delete agents for each node
+  kubernetes.core.k8s:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+    name: >-
+      {{
+        node_info.ports |
+        map(attribute='address') |
+        cloudkit.service.mac_to_agent_name(initial_agents)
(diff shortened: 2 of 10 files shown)


tzumainn (changes requested):
Minor comment: the `import_agents.yaml` and `remove_agents.yaml` playbooks can maybe be presented as agnostic to the ESI playbooks (although it currently isn't in practice). The `cluster_infra` roles are an existing example:

```
- name: Create cluster network
  ansible.builtin.include_role:
    name: massopencloud.esi.network
  vars:
    network_suffix: "{{ cluster_infra_name }}"
    network_state: present
    network_tags: ['o-sac', '{{ cluster_infra_name }}']
```

Right now they call the ESI playbooks directly; however the intent is that some time in the future, if/when we support something besides ESI, perhaps we call `{{ bare_metal_management_playbooks }}.network` instead; assuming that all the bare metal management playbooks take the same parameters, this `cluster_infra` playbook will just work against the new bare metal management playbooks.

What this means is that `import_agents.yaml` and `remove_agents.yaml` should maybe be taken out of the `massopencloud/esi` collection and be placed somewhere in `collections/ansible_collections/cloudkit/service` instead; and much of the ESI text in those two playbooks should be replaced to be more neutral ("Import nodes a agents", "node_names"). The one exception is `massopencloud.esi.agent`.

collections/ansible_collections/massopencloud/esi/roles/agent/tasks/import_agents.yaml
```diff
@@ -0,0 +1,186 @@
...
+  kubernetes.core.k8s_info:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ agent_namespace }}"
+  register: esi_initial_agents
+
+- name: Initialize esi_node_info_list and esi_node_addresses_list
+  ansible.builtin.set_fact:
+    esi_node_info_list: []
+    esi_node_addresses_list: []
```
tzumainn: I don't think you need this initialization, since you're setting the values directly below.

collections/ansible_collections/massopencloud/esi/roles/agent/tasks/import_agents.yaml
```diff
@@ -0,0 +1,186 @@
...
+  ansible.builtin.set_fact:
+    esi_node_addresses_list: "{{ port_list_raw.results | map(attribute='stdout') | map('from_json') | map('map', attribute='Address') | map('list') | list }}"
+
+- name: Initialize filtered lists of only nodes that need importing
+  ansible.builtin.set_fact:
+    filtered_node_names: []
+    filtered_node_info: []
+    filtered_node_addresses: []
+
+- name: Set filtered lists of only nodes that need importing
```
tzumainn: Could you simply match the node names against the agent's hostnames? I think that should be fine, since that's how you identify agents for removal.

tzumainn (changes requested):
I tested the playbooks - they worked well! Some more changes:

* `import_agents.yaml` and `remove_agents.yaml` won't ever be called independently; they should be put into  `collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks`, where they'll be called by your forthcoming playbooks

* Much of the non-ESI tasks in the ESI role should move into the cloudkit role (in `import_agents` and `remove_agents`). The general idea is that we want to avoid duplication of tasks if/when we add support for additional bare metal management platforms. So things like listing existing agents and determining which nodes need to be imported could move to the cloudkit role, while getting MAC addresses of nodes could stay in the ESI role. This may result in you creating multiple ESI roles for very specific, granular tasks (such as getting the MAC addresses of a node, or deploying a node with a given image).

* You should use the networking ESI roles where possible; for example, when you detach networks from the node and then attach them to the provisioning network, you can simply call `massopencloud.esi.l2.set_networks_for_node`.

collections/ansible_collections/massopencloud/esi/roles/agent/tasks/import_agents.yaml
ajamias: The reason why I create lists of filtered node information is to make the tasks concurrent so it doesn't take a lot of time to deploy one-by-one. There is also the `massopencloud.esi.match_agent_addresses` filter which I used that matches node addresses to agent addresses

tzumainn: Oh yeah, I agree that it's good to get a list of nodes you want to add; I'm just wondering why you need to match the ESI nodes against the agents by MAC address, rather than by node name.

Also, do you need both `filtered_node_names` and `filtered_node_info`? Wouldn't the latter include node name information?

ajamias:
This would go after the nodes have been removed as agents when the nodes are undeployed, cleaned, and provision_state is available?

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
```diff
@@ -0,0 +1,184 @@
...
+- name: Set deploy interface, set boot image url, and deploy node
+  openstack.cloud.baremetal_node_action:
+    name: "{{ node_info.name }}"
+    instance_info:
+      deploy_interface: ramdisk
+      boot_iso: "{{ discovery_url }}"
+    deploy: true
+  loop: "{{ filtered_node_info }}"
+  loop_control:
+    loop_var: node_info
```
tzumainn: `Set node to provisioning network` and this task could be combined into a single `massopencloud.esi` node deployment role; I would suggest something under `massopencloud/esi/roles/node`

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
```diff
@@ -0,0 +1,184 @@
...
+    namespace: "{{ manage_agents_namespace }}"
+  register: initial_agents
+
+- name: Get node informations
+  openstack.cloud.baremetal_node_info:
+    name: "{{ node_name }}"
+  register: node_infos
+  loop: "{{ manage_agents_node_names }}"
+  loop_control:
+    loop_var: node_name
```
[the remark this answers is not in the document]
tzumainn: Okay! I'd still have it call an ESI role if possible.

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
```diff
@@ -0,0 +1,135 @@
...
+  when: not discovery_url|default(false)
+
+- name: Provision nodes
+  ansible.builtin.include_role:
+    name: massopencloud.esi.node
+    tasks_from: provision_nodes
+  vars:
+    provisioning_network_name: "{{ manage_agents_provisioning_network_name }}"
+    node_info_list: "{{ filtered_node_info }}"
+    wait: true
```
tzumainn: I would make the image name a var rather than rely on `discovery_url` being set; it becomes much easier to reuse `provision_nodes` that way.

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
```diff
@@ -0,0 +1,135 @@
...
+  ansible.builtin.debug:
+    var: manage_agents_node_names
+
+- name: Get list of node info
+  ansible.builtin.include_role:
+    name: massopencloud.esi.node
+    tasks_from: list_nodes
+  vars:
+    node_names: "{{ manage_agents_node_names }}"
+# registers: node_info_list
```
tzumainn: May as well remove this!

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
```diff
@@ -0,0 +1,135 @@
...
+
+- name: Set filtered lists of only nodes that need importing
+  ansible.builtin.include_role:
+    name: massopencloud.esi.node
+    tasks_from: get_non_agent_nodes
+# registers: filtered_node_info
+# registers: agents
```
tzumainn: (and this)

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
```diff
@@ -0,0 +1,135 @@
...
+    wait: true
+
+- name: Get filtered agent names
+  ansible.builtin.include_role:
+    name: massopencloud.esi.node
+    tasks_from: get_agent_names
+  vars:
+    node_info_list: "{{ filtered_node_info }}"
+# registers: agent_names
+# registers: agent_hostnames
```
tzumainn: (and these, as well as the similar commented out text below)

collections/ansible_collections/massopencloud/esi/roles/node/tasks/get_agent_names.yaml
```diff
@@ -0,0 +1,34 @@
+- name: Get agents
+  kubernetes.core.k8s_info:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+  register: agents
```
tzumainn: I feel like this task should be done in the cloudkit playbook, and passed into here.

collections/ansible_collections/massopencloud/esi/roles/node/tasks/get_non_agent_nodes.yaml
```diff
@@ -0,0 +1,15 @@
+- name: Get agents
+  kubernetes.core.k8s_info:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+  register: agents
```
tzumainn: Same here

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
tzumainn: Is `node_info_list` used anywhere other than the task below? If not, I kind of think that it might belong within the `get_non_agent_nodes` task below.

collections/ansible_collections/massopencloud/esi/roles/node/tasks/get_agent_names.yaml
```diff
@@ -0,0 +1,34 @@
...
+  ansible.builtin.set_fact:
+    agent_hostnames: >-
+      {{
+        agent_hostnames | default([]) + [node_info.name | lower]
+      }}
+  loop: "{{ node_info_list }}"
+  loop_control:
+    loop_var: node_info
+    label: "{{ node_info.name }}"
+  when: node_info.ports | map(attribute='address') | massopencloud.esi.match_agent_addresses(agents.resources)
```
tzumainn: These two tasks loop the same way; is it possible to combine them and just return list of dicts?

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/remove_agents.yaml
```diff
@@ -0,0 +1,57 @@
+- name: Display node names that will be removed from being agents
+  ansible.builtin.debug:
+    var: manage_agents_node_names
+
+- name: Get list of existing agents
+  kubernetes.core.k8s_info:
+    kind: Agent
+    api_version: agent-install.openshift.io/v1beta1
+    namespace: "{{ manage_agents_namespace }}"
+  register: initial_agents
```
tzumainn: I don't think `initial_agents` is used, is it? Regardless, `get_agent_names` could remove its own task fetching agents, and just have this value passed in.

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/remove_agents.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+- name: Get list of node info
+  ansible.builtin.include_role:
+    name: massopencloud.esi.node
+    tasks_from: list_nodes
+  vars:
+    node_names: "{{ manage_agents_node_names }}"
```
tzumainn: I'd combine this was the `get_agent_names` role

collections/ansible_collections/cloudkit/service/roles/manage_agents/tasks/import_agents.yaml
```diff
@@ -0,0 +1,135 @@
...
+            (topology_labels_list[node_idx] |
+            from_yaml) |
+            combine(agent_resource_class_labels_list[node_idx])
+          }}
+      spec:
+        approved: true
+        hostname: "{{ filtered_node_info[node_idx].name | lower }}"
+  loop: "{{ range(agent_names | length) | list }}"
+  loop_control:
+    loop_var: node_idx
```
tzumainn: I think I'd have a single esi role that encompasses the setting of labels on a list of agents (so a role that extracts the topology and does the other things before setting the labels). Apparently different bare metal management systems may end up with a different set of labels, and what we're doing here is likely very ESI specific.

tzumainn (changes requested):
Some additional comments!

tzumainn (changes requested):
I think this is almost there! I've put some inline suggestions, and have one more general suggestion: change `clean_nodes.yaml` and `provision_nodes.yaml` to act on a single node; and in the cloudkit playbooks, get the list of nodes first, and then simply loop over them calling `clean` or `provision` multiple times.

PR #158, opened 2025-11-05
Add stale ESI resources cleanup playbooks

DanNiESh:
This cleanup script runs per instance to identify and remove stale network resources (networks, floating IPs, routers, subnets) that are no longer associated with existing cluster orders

- The cleanup playbook filters resources by instance and purpose tags, then removes those that don't have a tag matching an existing ClusterOrder.

This PR closes [innabox #189](https://github.com/innabox/issues/issues/189)

### DIFF
--- a/collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/defaults/main.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/defaults/main.yaml
@@ -1,0 +1,1 @@
+cleanup_stale_network_resources_namespace: "{{ lookup('env', 'POD_NAMESPACE') }}"
--- a/collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/meta/argument_specs.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/meta/argument_specs.yaml
@@ -1,0 +1,8 @@
+argument_specs:
+  main:
+    options:
+      cleanup_stale_network_resources_namespace:
+        description: |
+          Namespace to filter network resources
+        type: str
+        required: true
--- a/collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
@@ -1,0 +1,166 @@
+---
+- name: Get all existing ClusterOrders from specified namespace
+  kubernetes.core.k8s_info:
+    api_version: cloudkit.openshift.io/v1alpha1
+    kind: ClusterOrder
+    namespace: "{{ cleanup_stale_network_resources_namespace }}"
+  register: cluster_orders
+  failed_when: false
+
+- name: Set cluster order names
+  ansible.builtin.set_fact:
+    existing_cluster_order_names: "{{ [prefix] | product(cluster_names) | map('join') }}"
+  vars:
+    prefix: "clusterorder_"
+    cluster_names: "{{ cluster_orders.resources | map(attribute='metadata.name') }}"
+
+- name: Display existing cluster names
+  ansible.builtin.debug:
+    msg: |
+      Found existing ClusterOrders: {{ existing_cluster_order_names }}
+
+# Networks
+- name: Get all networks with purpose_o-sac and instance tags
+  ansible.builtin.include_role:
+    name: massopencloud.esi.l2
+    tasks_from: list_networks
+  vars:
+    l2_network_properties:
+      purpose: "o-sac"
+      instance: "{{ cleanup_stale_network_resources_namespace }}"
+
+- name: Get stale networks
+  ansible.builtin.set_fact:
+    cleanup_stale_network_resources_stale_networks: "{{ cleanup_stale_network_resources_stale_networks + [item] }}"
+  loop: "{{ list_networks_result | default([]) }}"
+  when:
+    - item.Tags is defined
+    - ((item.Tags | default([])) | intersect(existing_cluster_order_names)) | length == 0
+
+# Floating IPs
+- name: Get all floating IPs with o-sac and instance tags
+  ansible.builtin.include_role:
+    name: massopencloud.esi.floating_ip
+    tasks_from: list_floating_ips
+  vars:
+    floating_ip_properties:
+      purpose: "o-sac"
+      instance: "{{ cleanup_stale_network_resources_namespace }}"
+
+- name: Get stale floating IPs
+  ansible.builtin.set_fact:
+    cleanup_stale_network_resources_stale_floating_ips: "{{ cleanup_stale_network_resources_stale_floating_ips + [item] }}"
+  loop: "{{ list_floating_ips_result | default([]) }}"
+  when:
+    - item.Tags is defined
+    - ((item.Tags | default([])) | intersect(existing_cluster_order_names)) | length == 0
+
+# Routers
+- name: Get all routers with o-sac and instance tags
+  ansible.builtin.include_role:
+    name: massopencloud.esi.l3
+    tasks_from: list_routers
+  vars:
+    l3_network_properties:
+      purpose: "o-sac"
+      instance: "{{ cleanup_stale_network_resources_namespace }}"
+
+- name: Get stale routers
+  ansible.builtin.set_fact:
+    cleanup_stale_network_resources_stale_routers: "{{ cleanup_stale_network_resources_stale_routers + [item] }}"
+  loop: "{{ list_routers_result | default([]) }}"
+  when:
+    - item.Tags is defined
+    - ((item.Tags | default([])) | intersect(existing_cluster_order_names)) | length == 0
+
+# Subnets
+- name: Get all subnets with o-sac and instance tags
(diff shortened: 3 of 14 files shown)


collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/meta/argument_specs.yaml
```diff
@@ -0,0 +1,11 @@
+argument_specs:
+  main:
+    options:
+      cleanup_stale_network_resources_osac_tag:
+        type: str
+        required: true
+      cleanup_stale_network_resources_instance_name:
```
larsks: I think this should be called `cleanup_stale_network_resources_namespace`, since we're explicitly using it for that purpose in e.g. the "Get all existing ClusterOrders" task.

collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
```diff
@@ -0,0 +1,256 @@
...
+  ansible.builtin.debug:
+    msg: |
+      Found existing ClusterOrders: {{ existing_cluster_names }}
+
+- name: Initialize stale resources lists
+  ansible.builtin.set_fact:
+    stale_networks: []
+    stale_floating_ips: []
+    stale_routers: []
+    stale_subnets: []
```
larsks: These could be set in `vars/main.yaml` instead, since we're initializing them with static values.

tzumainn (commented):
Would it be possible to split this PR - one for updating the tags, and one for the cleanup_stale_network_resources stuff?

I think I might rename the `*_attributes` to something else; to me, "attributes" implies that we're setting something directly on the object, as opposed to adding arbitrary key value pairs. Ironic uses `property`, so maybe that?

collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
```diff
@@ -0,0 +1,256 @@
...
+
+- name: Initialize existing cluster names list
+  ansible.builtin.set_fact:
+    existing_cluster_names: []
+
+- name: Extract cluster names from ClusterOrders
+  ansible.builtin.set_fact:
+    existing_cluster_names: "{{ existing_cluster_names + ['clusterorder_' + item.metadata.name] }}"
+  loop: "{{ cluster_orders.resources }}"
+  when: cluster_orders.resources is defined
```
larsks: You don't need a separate task for variable initialization; you can add a `vars:` clause to the task instead:

```suggestion
- name: Extract cluster names from ClusterOrders
  ansible.builtin.set_fact:
    existing_cluster_names: "{{ existing_cluster_names + ['clusterorder_' + item.metadata.name] }}"
  loop: "{{ cluster_orders.resources }}"
  when: cluster_orders.resources is defined
  vars:
    existing_cluster_names: []
```

collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
```diff
@@ -0,0 +1,245 @@
...
+  register: cluster_orders
+  failed_when: false
+
+- name: Initialize existing cluster names list
+  ansible.builtin.set_fact:
+    existing_cluster_names: []
+
+- name: Extract cluster names from ClusterOrders
+  ansible.builtin.set_fact:
+    existing_cluster_names: "{{ existing_cluster_names + ['clusterorder_' + item.metadata.name] }}"
```
tzumainn: I think you should be able to avoid the initialization task above with something like `existing_cluster_names: "{{ existing_cluster_names | default([]) + ['clusterorder_' + item.metadata.name] }}"`

collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
```diff
@@ -0,0 +1,245 @@
...
+    msg: |
+      Found existing ClusterOrders: {{ existing_cluster_names }}
+
+# Networks
+- name: Get all networks with purpose_o-sac and instance tags
+  ansible.builtin.command: >-
+    openstack network list
+    --tag {{ cleanup_stale_network_resources_osac_tag }}
+    --tag instance_{{ cleanup_stale_network_resources_namespace }}
+    -f json
```
tzumainn: We don't want to call `openstack` directly in any of the cloudkit playbooks (if you check, you'll see we explicitly avoid doing so). Instead, you'll want to create a new `massopencloud.esi.l2` task that does this. I'd suggest having that task take in a `property` parameter, similar to what was used when creating the network, and then having the task translate that into a tag.

tzumainn (changes requested):
Some initial feedback!

collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
DanNiESh: I see! When the cluster_orders.resource is [], 
```
  tasks:
    - name: Extract cluster names from ClusterOrders
      ansible.builtin.set_fact:
        existing_cluster_names: "{{ existing_cluster_names + ['clusterorder_' + item.metadata.name] }}"
      loop: "{{ cluster_orders.resources }}"
      when: cluster_orders.resources is defined
      vars:
        existing_cluster_names: []
``` 
this won't get executed, so existing_cluster_names is not defined. How should we handle this case?

larsks: You're already defaulting several other variables to empty lists (`[]`); you could handle `existing_cluster_names` in the same place. I think I commented somewhere else that you could do this in `vars/main.yaml` or `defaults/main.yaml`.

collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
```diff
@@ -0,0 +1,245 @@
...
+  register: networks_with_tag
+  changed_when: false
+  failed_when: false
+
+- name: Parse networks JSON
+  ansible.builtin.set_fact:
+    networks_json: "{{ networks_with_tag.stdout | from_json | default([]) }}"
+  when: networks_with_tag.stdout is defined
+
+- name: Get detailed network information with tags
```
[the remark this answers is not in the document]
DanNiESh: Done!

collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/defaults/main.yaml
```diff
@@ -0,0 +1,2 @@
+cleanup_stale_network_resources_osac_tag: 'purpose_o-sac'
```
tzumainn: I think this can be removed (let me know if I missed where it's used!)

collections/ansible_collections/cloudkit/service/roles/cleanup_stale_network_resources/tasks/cleanup.yaml
```diff
@@ -0,0 +1,165 @@
...
+    kind: ClusterOrder
+    namespace: "{{ cleanup_stale_network_resources_namespace }}"
+  register: cluster_orders
+  failed_when: false
+
+- name: Extract cluster names from ClusterOrders
+  ansible.builtin.set_fact:
+    cleanup_stale_network_resources_cluster_orders: "{{ cleanup_stale_network_resources_cluster_orders + [item] }}"
+  loop: "{{ cluster_orders.resources }}"
+  when: cluster_orders.resources is defined
```
tzumainn: It's my impression that `cluster_orders.resources` will always be defined; if there are no matching resources, then it'll simply be an empty list. Have you seen something different?

Also, if this is looking for the cluster order names, I think you need something more like:

```
    - name: set cluster order names
      ansible.builtin.set_fact:
        cluster_order_names: "{{ [prefix] | product(cluster_names) | map('join') }}"
      vars:
        prefix: clusterorder_
	    cluster_names: "{{ cluster_orders.resources | map(attribute='metadata.name') }}"
```

tzumainn (approved):
After directly testing the role, I'm good with this!

@larsks  I'm a little fuzzy on how these playbooks make it into AAP - besides the top-level playbook, do they also have to be added to `collections/ansible_collections/cloudkit/config_as_code/roles/aap/vars/controller.yml`?

larsks (approved):
Thanks!

larsks:
> @larsks I'm a little fuzzy on how these playbooks make it into AAP - besides the top-level playbook, do they also have to be added to collections/ansible_collections/cloudkit/config_as_code/roles/aap/vars/controller.yml?

**IF** we want to create AAP objects (job templates, workflow templates) to execute these resources, **THEN** we need to update resources in the config_as_code collection. That contains everything responsible for configuring AAP.

I asked about this [earlier](https://github.com/innabox/cloudkit-aap/pull/158#pullrequestreview-3423027063) but never received a reply. @DanNiESh, pinging you on that question; if the answer is "no" we should create a new issue.

PR #159, opened 2025-11-05
Update network resources tags

DanNiESh:
This is part of [innabox #189](https://github.com/innabox/issues/issues/189)

- CloudKit playbooks create network/floating IP resources with key-value pairs in the format:
```
{"purpose": "o-sac",
 "instance": "<instance_name>,
 "clusteroder": <cluster_order_name>}
```
- ESI playbooks convert the key value pairs to tag list `['purpose_o-sac', 'instance_<name>', 'clusterorder_<cluster order name>']` and apply to network resources

### DIFF
--- a/collections/ansible_collections/cloudkit/service/roles/cluster_infra/tasks/create_cluster_infra.yaml
+++ b/collections/ansible_collections/cloudkit/service/roles/cluster_infra/tasks/create_cluster_infra.yaml
@@ -8,7 +8,10 @@
   vars:
     network_suffix: "{{ cluster_infra_name }}"
     network_state: present
-    network_tags: ['o-sac', '{{ cluster_infra_name }}']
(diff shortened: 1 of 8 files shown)


tzumainn (changes requested):
Testing it out, the following tags were generated:

`clusterorder_order-xcr4b, instance_cluster-order-xcr4b-6pdpkh, purpose_o-sac`

The instance should be `tzumainn` as the namespace in which my instance is installed; the `cluster_working_namespace` is the namespace for the tenant cluster itself. I think what you may want is `cluster_order.metadata.namespace` instead

tzumainn:
Perfect! These are the tags that are now added to a network:

```
| tags                      | clusterorder_order-cfvnx, instance_tzumainn, purpose_o-sac |
```

PR #160, opened 2025-11-05
Allocate Floating IP for VM

ajamias:
This adds the ability to ssh a floating IP allocated by OpenStack into a VM. The floating IP is automatically allocated when the playbook to create a vm is ran, and automatically removed when the playbook to delete a vm is ran. Resolves [innabox/#241](https://github.com/orgs/innabox/projects/13/views/1?pane=issue&itemId=133810235&issue=innabox%7Cissues%7C241)

### DIFF
--- a/collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/tasks/create.yaml
+++ b/collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/tasks/create.yaml
@@ -8,8 +8,7 @@
     vm_image_source: "{{ template_parameters.vm_image_source | default(default_vm_image_source) }}"
     vm_os_type: "{{ template_parameters.vm_os_type | default(default_vm_os_type) }}"
     storage_class: "{{ default_vm_storage_class }}"
-    vm_expose_service: "{{ default_vm_expose_service }}"
-    vm_service_type: "{{ default_vm_service_type }}"
+    vm_internal_network: "{{ default_vm_internal_network }}"
     vm_service_ports: "{{ default_vm_service_ports }}"
     vm_network_type: "{{ template_parameters.vm_network_type | default(default_vm_network_type) }}"
     ssh_public_key: "{{ template_parameters.ssh_public_key | default(default_ssh_public_key) }}"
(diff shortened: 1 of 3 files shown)


collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/tasks/create.yaml
```diff
@@ -143,6 +143,90 @@
...
+  ansible.builtin.include_role:
+    name: massopencloud.esi.floating_ip
+    tasks_from: create_port_forwarding
+  vars:
+    internal_ip: "{{ node_info.resources[0].status.addresses | selectattr('type', 'equalto', 'InternalIP') | map(attribute='address') | first }}"
+    floating_ip: "{{ allocate_floating_ip_result }}"
+    internal_network: hypershift
+    ports:
+      - "22:{{ load_balancer.resources[0].spec.ports | selectattr('name', 'equalto', 'ssh') | map(attribute='nodePort') | first }}"
+      - "80:{{ load_balancer.resources[0].spec.ports | selectattr('name', 'equalto', 'http') | map(attribute='nodePort') | first }}"
```
adriengentil: You we try to map them on `vm_service_ports` ?

adriengentil:
> Should the floating ip not be allocated if ssh (port 22) isn't one of the vm_service_ports?

it might be weird in terms of UX for the user, that would mean the IP can change if they update the ports?

ajamias:
> it might be weird in terms of UX for the user, that would mean the IP can change if they update the ports?

I'll leave it as it is then

collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/tasks/create.yaml
```diff
@@ -127,44 +127,32 @@
...
 - name: Create a floating IP to attach to VM
   ansible.builtin.include_role:
     name: massopencloud.esi.floating_ip
   vars:
     floating_ip_state: present
     floating_ip_name: "{{ vm_name }}-floating-ip"
     floating_ip_network: external
+    floating_ip_properties:
+      vm_name: "{{ vm_name }}"
+      vm_namespace: "{{ vm_target_namespace }}"
```
tzumainn: Maybe just call it `vm` instead of `vm_name`.

I would say we also want `instance` instead of `vm_namespace`. `instance` refers to the particular installation of O-SAC; for example, right now in hypershift1 we have the `ajamais` instance and the `tzumainn` instance, as well as `innabox-isaiah` and `innabox-lars`. The goal is to distinguish between these instances.

I'm also not sure we want to set `instance` to `vm_target_namespace`, which I *think* is a new namespace specifically created to hold VM resources. We would want something like `vm_order.metadata.namespace`. If that works, then in your testing you should see a floating IP with a tag of `instance_ajamais`

collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/tasks/create.yaml
ajamias: I map to the vm's node's IP:NodePort, and the NodePorts are dynamically created from the vm_service_ports

collections/ansible_collections/cloudkit/templates/roles/ocp_virt_vm/tasks/create.yaml
```diff
@@ -127,21 +127,91 @@
...
-  when: vm_expose_service and vm_service_ports | length > 0
+    definition:
+      metadata:
+        annotations:
+          cloudkit.openshift.io/floating-ip-address: "{{ allocate_floating_ip_result }}"
+          cloudkit.openshift.io/floating-ip-name: "{{ vm_name }}-floating-ip"
+          cloudkit.openshift.io/load-balancer-ip: "{{ load_balancer.resources[0].status.loadBalancer.ingress[0].ip }}"
+          cloudkit.openshift.io/vm-node-name: "{{ vmi_info.resources[0].status.nodeName }}"
+          cloudkit.openshift.io/vm-node-ip: "{{ node_info.resources[0].status.addresses | selectattr('type', 'equalto', 'InternalIP') | map(attribute='address') | first }}"
+          cloudkit.openshift.io/floating-ip-network: "{{ floating_ip_network | default('external') }}"
```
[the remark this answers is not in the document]
adriengentil: no, my suggestion is to keep `cloudkit.openshift.io/` for stuff that we consider generic, and that can be referenced in `cloudkit-operator`, and use another domain/namespace for stuff specific to playbooks/underlying infra.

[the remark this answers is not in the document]
tzumainn: Ah, got it! I do agree. To be honest, I kind of think that most of these annotations could be removed, other than `public_ip_address`. I know that `cloudkit-aap.openshift.io/floating-ip-network` is used in the cleanup playbook - but the cluster playbooks simply use the default external network specified in the ESI role, and I think these playbooks should do the same.

tzumainn (commented):
Two last minor comments; then it looks good to me from an ESI perspective!

PR #169, opened 2025-11-14
Run post-delete job to report status

larsks:
This replaces the "delete-hosted-cluster" job with a workflow
("delete-hosted-cluster-workflow") that runs a status reporting job after
the main job has finished. The status jobs are responsible for updating the
clusterorder with the success or failure of the delete operation.

This can be useful both to expose the status to the operator (from where we
could in theory publish it to the fulfillment service) and to possibly
re-trigger the delete webhook.

Currently the status is reported as an annotation on the clusterorder,
rather than as a status condition, because the kubernetes ansible modules
do not permit updates to the status subresource of kubernetes objects. See
[1] for details.

[1]: https://github.com/ansible-collections/kubernetes.core/issues/607

Part of: https://github.com/innabox/issues/issues/239

### DIFF
--- a/playbook_cloudkit_create_hosted_cluster.yml
+++ b/playbook_cloudkit_create_hosted_cluster.yml
@@ -49,17 +49,25 @@
         lease_holder: "{{ cluster_order_holder_id }}"
         lease_name: "cluster-{{ cluster_order_name }}-lock"
 
-    - name: Add infrastructure finalizer to clusterorder
+    - name: Add infrastructure finalizer
       ansible.builtin.include_role:
         name: cloudkit.service.finalizer
       vars:
         finalizer_state: present
         finalizer_name: "{{ cluster_order_infrastructure_finalizer }}"
         finalizer_target:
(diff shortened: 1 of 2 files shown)


tzumainn (commented):
This makes a lot of sense; is there any reason not to also add it to the `create-hosted-cluster-workflow`?

larsks:
> Actually, here's a more complete thought: what if we used this as a standard way to show the status of the cluster order specifically within the fulfillment service?

I think those are great enhancements for a future PR. I would prefer to first have a solution that allows us to update status conditions rather than using annotations.

- https://github.com/innabox/issues/issues/310
- https://github.com/innabox/issues/issues/311

larsks:
@jkary I've resolved the linting errors.

tzumainn:
> > the infrastructure finalizer is added to the namespace, because that is where the cluster order exists
> 
> No. The cluster order exists in the namespace in which our services are running.

Ah, got it; okay, makes sense. Thanks!

PR #173, opened 2025-11-17
Updated pre-commit to support ansible-lint

jkary:


.pre-commit-config.yaml
```diff
@@ -18,4 +18,15 @@ repos:
       files: \.(yaml|yml)$
       types: [file, yaml]
       entry: yamllint --strict
+
+- repo: https://github.com/ansible/ansible-lint
```
adriengentil: if we use pre-commit now, can you remove https://github.com/innabox/cloudkit-aap/blob/main/.github/workflows/tests.yml ?

larsks: To be clear, my suggestion was to *keep* `main/.github/workflows/tests.yml` and not put `ansible-lint` into the pre commit checks. I think it will obscure other lint results and would be more appropriate as a pre-push hook.

eranco74:
@jkary @larsks can we close this?

larsks:
Closed!

PR #190, opened 2025-12-12
Replace ESI CLI Commands With Ansible Modules

ajamias:
There are some roles in massopencloud.esi that depend on using the
python-esiclient cli commands to perform actions. This commit moves the
functionality into their own Ansible modules only relying on OpenStack's
openstacksdk.

This commit also replaces instances of `openstack esi node network` with a
call to the module

### DIFF
--- a/collections/ansible_collections/massopencloud/esi/plugins/modules/node_network.py
+++ b/collections/ansible_collections/massopencloud/esi/plugins/modules/node_network.py
@@ -1,0 +1,194 @@
+#!/usr/bin/python
+
+DOCUMENTATION = r'''
+---
+module: node_network
+short_description: Sets the networks attached to a baremetal node
+description:
+  - Manage Neutron networks attached to an Ironic node
+  - Sets the node's networks to exactly match the specified list
+  - This means it will detach any networks not in the list and attach any missing networks
+  - If networks is empty or not provided, detaches all networks from the node
+options:
+  networks:
+    description:
+      - List of network names or IDs to set on the node
+      - If empty or not provided, all networks will be detached
+    type: list
+    elements: str
+  node:
+    description:
+      - The name or ID of the node
+    required: true
+    type: str
+extends_documentation_fragment:
+  - openstack.cloud.openstack
+'''
+
+EXAMPLES = r'''
+- name: Set a single network on a node
+  massopencloud.esi.node_network:
+    cloud: "devstack"
+    node: "MOC-R4PAC67U12-S3"
+    networks:
+      - "hypershift"
+
+- name: Set node networks to exactly match the specified list
+  massopencloud.esi.node_network:
+    cloud: "devstack"
+    node: "MOC-R4PAC67U12-S3"
+    networks:
+      - "hypershift"
+      - "provisioning"
+      - "storage"
+
+- name: Detach all networks from a node (empty networks list)
+  massopencloud.esi.node_network:
+    cloud: "devstack"
+    node: "MOC-R4PAC67U12-S3"
+    networks: []
+
+- name: Detach all networks from a node (no networks parameter)
+  massopencloud.esi.node_network:
+    cloud: "devstack"
+    node: "MOC-R4PAC67U12-S3"
+'''
+
+from ansible_collections.openstack.cloud.plugins.module_utils.openstack import (
+        OpenStackModule
+)
+
+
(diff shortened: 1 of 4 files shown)


collections/ansible_collections/massopencloud/esi/plugins/modules/node_network.py
```diff
@@ -0,0 +1,198 @@
...
+    node: "MOC-R4PAC67U12-S3"
+    network: "hypershift"
+
+- name: Detach all networks from a node
+  massopencloud.esi.node_network:
+    cloud: "devstack"
+    state: "absent"
+    node: "MOC-R4PAC67U12-S3"
+'''
+
```
tzumainn: In my opinion, it'd make more sense to have the module take an array of networks, and ensure that the node networking is set to exactly those networks. I know that's what the role does, but I think it's fine for the role to simply call this module directly without any additional bells or whistles.

tzumainn (changes requested):
`node_network_info` is (understandably) a mess, but your PR made me realize that we don't actually use `get_node_network_info` anywhere. Do you think it'd be possible to simply get rid of it, and just use simplified logic in `node_network.py` to ensure node networks are set as the user desires?

collections/ansible_collections/massopencloud/esi/plugins/modules/node_network.py
tzumainn: In my opinion, yes; OSAC currently only deals with networks, and my guess is that when we get to trunks, we'll be passing in networks into the module and having the module create the trunk.

tzumainn (commented):
Tested this out on a hostpool, and things seem to work!

I do have one opinion/question: I think the ansible module shouldn't take a `present` or `absent`; rather, it should just take the node name and the list of networks, and then set the networks to *exactly* the array of networks. To me, that most closely matches ansible's "assertive" nature: if we assert that node should be attached to networks A and B, then it should be attached to exactly A and B, and nothing else.

@larsks you're the expert here - am I right, or not-so-right?

tzumainn:
> Tested this out on a hostpool, and things seem to work!
> 
> I do have one opinion/question: I think the ansible module shouldn't take a `present` or `absent`; rather, it should just take the node name and the list of networks, and then set the networks to _exactly_ the array of networks. To me, that most closely matches ansible's "assertive" nature: if we assert that node should be attached to networks A and B, then it should be attached to exactly A and B, and nothing else.
> 
> @larsks you're the expert here - am I right, or not-so-right?

@ajamias I talked with lars and he agreed with me; basically, the idea is that when we call this module to attach networks A and B to a node, we want exactly A and B - and no other networks. The model I'm suggesting would handle that, while the current one has the potential to leave old networks attached to the node.

collections/ansible_collections/massopencloud/esi/roles/l2/tasks/set_networks_for_node.yaml
```diff
@@ -22,28 +12,13 @@
...
-  loop: "{{ nnl_result }}"
-  loop_control:
-    loop_var: network_info
-  openstack.cloud.port:
-    state: "absent"
-    name: "{{ network_info['Network Port UUID'] }}"
+- name: Detach networks from node
+  massopencloud.esi.node_network:
+    state: absent
+    node: "{{ node }}"
```
tzumainn: We shouldn't need this anymore, right?

tzumainn (changes requested):
I had trouble attaching two networks to a node; the first network would attach properly, and then the second would complain:

```
Failed to attach VIF bd040b7e-355\
a-4752-9fd9-eddeeccd1cdf to bare metal node a1626898-d5c6-4312-b5aa-385bdaf9e4cc: Client Error for url: https://esi.massopen.cloud:13385/v1/nodes/a1626898-d5c6-4312-b5aa-385bdaf9e4cc/vifs, Unable to attach VIF bd040b7e-355a-4752-9fd9-eddeeccd1cdf, not enough free physical ports.
```

However, I was able to run `openstack esi node network attach` directly to attach the second network.

eranco74:
@ajamias  Is this still relevant? 
Should we close this?

ajamias:
> @ajamias Is this still relevant? Should we close this?

I think this will get attention after we demo bare metal fulfillment in CasS.

PR #234, opened 2026-04-07
MGMT-23770: fix publish_templates for empty API responses

zszabo-rh:
## Summary

- Fix `publish_templates` role failing when the fulfillment-service API returns empty responses (no `size` field)
- Replace `.json.size > 0` with `.json.get('items', []) | length > 0` in all 3 task files
- Add unit tests with a mock HTTP server covering empty, populated, and edge-case responses

## Root Cause

The fulfillment-service uses proto3 `optional int32` for the `size` field. When there are no items, proto3 JSON serialization omits zero-valued optional fields entirely, so the response is `{}` instead of `{"size": 0, "items": []}`. The Ansible conditional `.json.size` then fails with `'dict object' has no attribute 'size'`.

The fix uses `.json.get('items', [])` (Python dict method) rather than dot or bracket notation, because `items` is also a built-in dict method in Python — both `.json.items` and `.json['items']` resolve to `dict.items()` when the key is absent.

## Files Changed

**Bug fix** (3 files, 1 line each):
- `collections/.../publish_templates/tasks/clusters.yaml`
- `collections/.../publish_templates/tasks/compute_instances.yaml`
- `collections/.../publish_templates/tasks/network_classes.yaml`

**Unit tests** (2 new files):
- `collections/.../publish_templates/tests/test.yml` — 3 test scenarios
- `collections/.../publish_templates/tests/mock_api_server.py` — mock HTTP server

## Test plan

- [x] `ansible-lint` passes (0 failures, 0 warnings)
- [x] `pre-commit run --all-files` passes
- [x] Unit test: empty API response (the [MGMT-23770](https://redhat.atlassian.net/browse/MGMT-23770) bug scenario)
- [x] Unit test: populated response (existing items PATCHed, new POSTed)
- [x] Unit test: no `items` key edge case
- [x] E2E on hypershift1 (osac-zszabo): publish_templates playbook against real fulfillment-service with empty network_classes and compute_instance_templates endpoints

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved template publication checks to reliably detect existing items across varied API JSON shapes, ensuring create vs update actions run correctly for clusters, compute instances, and network classes.

* **Tests**
  * Added an end-to-end test playbook and a local mock API server to validate populated, empty, and missing-items responses and expected POST vs PATCH behavior.

### DIFF
--- a/collections/ansible_collections/osac/service/roles/publish_templates/tests/test.yml
+++ b/collections/ansible_collections/osac/service/roles/publish_templates/tests/test.yml
@@ -1,0 +1,255 @@
+---
+# Test playbook for osac.service.publish_templates role.
+# Each play is guarded by a `when` condition with bool coercion -- only the
+# play whose variable evaluates to true will run. Pass exactly one variable
+# per invocation.
+#
+# These tests use a Python mock HTTP server to simulate the fulfillment-service
+# API, so they can run without any external dependencies.
+#
+# Required variables for every test:
+#   test_mock_port:  Port for the mock HTTP server (default: 18081)
+#
+#   Test 1 (empty API responses -- the MGMT-23770 bug scenario):
+#     When no items exist, proto3 omits the `size` field entirely.
+#     The role must handle {"items": []} without a `size` key.
+#     Expected: All templates are created via POST, no PATCH calls.
+#     ansible-playbook ... -e test_empty=true
+#
+#   Test 2 (populated API responses -- existing items):
+#     When items already exist, the role should PATCH them.
+#     New items that don't exist should be POSTed.
+#     Expected: Existing templates updated via PATCH, new ones POSTed.
+#     ansible-playbook ... -e test_populated=true
+#
+#   Test 3 (no items key in response -- edge case):
+#     Some API responses might return {} without an items key.
+#     The role should treat this as empty and POST all templates.
+#     Expected: All templates created via POST.
+#     ansible-playbook ... -e test_no_items_key=true
+
+# ──────────────────────────────────────────────────────────────
+# Test 1: Empty API responses (the MGMT-23770 bug scenario)
+# Expected: role succeeds, all templates created via POST
+# ──────────────────────────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


collections/ansible_collections/osac/service/roles/publish_templates/tests/test.yml
```diff
@@ -0,0 +1,251 @@
...
+        url: "http://127.0.0.1:{{ test_mock_port }}/_calls"
+        return_content: true
+      register: call_log
+      when: test_populated is defined
+
+    - name: Verify PATCH calls for existing items and POST for new
+      ansible.builtin.assert:
+        that:
+          - call_log.json | selectattr('method', 'equalto', 'PATCH') | list | length == 3
+          - call_log.json | selectattr('method', 'equalto', 'POST') | list | length == 2
```
akshaynadkarni: **(from claude)**, you can only include if you feel this is necessary. 



*Suggestion for stronger assertions* 

The populated test verifies the correct number of PATCH and POST calls, which is good. To strengthen confidence, you could also verify that PATCHes went to the right endpoints (e.g., that the cluster PATCH included the existing ID in the URL path). This would catch a regression where the role PATCHes the wrong endpoint or mixes up resource types.

For example:
```yaml
  - call_log.json | selectattr('method', 'equalto', 'PATCH') | selectattr('path', 'contains', 'cluster_templates') | list | length == 1
  - call_log.json | selectattr('method', 'equalto', 'PATCH') | selectattr('path', 'contains', 'compute_instance_templates') | list | length == 1
  - call_log.json | selectattr('method', 'equalto', 'PATCH') | selectattr('path', 'contains', 'network_classes') | list | length == 1
```

akshaynadkarni (approved):
Code LGTM. I've left a few comments/suggestions based on my review using Claude's help. None of these are blockers. You can decide whether/not to address them. 

Thanks for adding the tests 🙏 . 

You are good to go once you address the checks.

akshaynadkarni (approved):
LGTM

PR #239, opened 2026-04-10
MGMT-23822: add unit tests for osac.service roles

zszabo-rh:
## Summary

- Add unit tests for 7 osac.service roles (up from 2 to 9 roles with test coverage)
- Two tiers: offline roles (no cluster needed) and kind-backed roles (need K8s API)
- 22 test scenarios total, all passing
- Fix missing `osac_network_classes` and `publish_templates_network_class_api_endpoint` in publish_templates argument_specs

## Tier 1: Offline roles (no cluster needed)

| Role | Scenarios | What's tested |
|------|-----------|---------------|
| `enumerate_templates` | 4 | Template discovery from collections: happy path, nonexistent collection, invalid name, mixed valid/invalid |
| `extract_template_info` | 4 | ClusterOrder spec parsing: qualified ID, unqualified ID prefix, pre-defined preservation, empty params |
| `common` | 2 | Remote kubeconfig env var: set and unset |

## Tier 2: Kind-backed roles (need K8s API, kind cluster sufficient)

| Role | Scenarios | What's tested |
|------|-----------|---------------|
| `finalizer` | 4 | Add/remove finalizers on ConfigMap: add, remove, idempotent add, remove nonexistent |
| `lease` | 2 | Coordination Lease lifecycle: acquire+release, release nonexistent |
| `cluster_working_namespace` | 3 | Namespace lookup by label: found, not found (block/rescue), pre-defined |
| `compute_instance_working_namespace` | 3 | Tenant CR lookup: found (status subresource), not found (block/rescue), pre-defined |

## Minor fix: publish_templates argument_specs

Added missing `osac_network_classes` and `publish_templates_network_class_api_endpoint` options to `publish_templates/meta/argument_specs.yaml`. These were added to `defaults/main.yaml` in commit 977c891 but not to the argument specs. Addresses review feedback from PR #234.

## Test conventions (matching existing patterns)

- Each role's tests in `roles/<name>/tests/test.yml`
- One variable per scenario, passed via `-e` (e.g., `-e test_add_finalizer=true`)
- Guards via `when: test_var | default(false) | bool`
- Error-path tests use `block/rescue` with `ansible.builtin.assert`
- All tests pass `ansible-lint`

## How to run

```bash
# Tier 1 (no setup needed):
ansible-playbook .../enumerate_templates/tests/test.yml -e test_discover_all=true

# Tier 2 (need kind cluster):
kind create cluster --name osac-role-tests
kubectl apply -f /path/to/osac-operator/config/crd/bases/osac.openshift.io_tenants.yaml
ansible-playbook .../finalizer/tests/test.yml -e test_add_finalizer=true
```

## Test plan

- [x] All 22 test scenarios pass locally
- [x] `ansible-lint` passes on all roles (0 failures, 0 warnings)
- [x] Tier 2 tests verified against kind cluster with OSAC CRDs

## Summary by CodeRabbit

* **Tests**
  * Added new unit and integration test playbooks for multiple roles, covering manifest/template extraction, template enumeration, namespace/tenant resolution, lease lifecycle, and finalizer behavior; includes role-level baseline scenarios and environment-driven cases.

* **Chores**
  * Improved test infrastructure: dedicated test-runner pod UID handling and orchestration to run workflow and role-level tests.
  * Added network-classes and network-class API endpoint configuration options to template publication.


zszabo-rh:
Addressed CodeRabbit's actionable feedback in a46f0a6:

- **block/always cleanup**: All Tier 2 tests now wrap test logic in `block:` with cleanup in `always:`, so resources are cleaned up even when assertions fail.
- **Lease weak assertion**: Replaced `that: true` with a `k8s_info` check confirming the lease doesn't exist.

**Not changing** (intentional design choices):

- **Hard-coded template IDs in enumerate_templates**: These are regression tests — they verify that _specific_ known templates (`ocp_4_17_small`, `ocp_virt_vm`, `cudn_net`) are discovered correctly. Weakening to `length >= 1` would miss cases where a template stops being discovered due to metadata or filter changes. The templates are part of the same repo, so coupling is acceptable.

- **Test selector validation in common**: Over-engineering for a 2-test playbook. The `when` guards already ensure only the selected test runs; misconfigured selectors simply result in all tasks being skipped, which is visible in the PLAY RECAP.

trewest:
@zszabo-rh Is there a reason not to add the tier 2 tests to the integration tests?

zszabo-rh:
@trewest 
Actually there's no particular reason I put these as standalone role tests rather than integration tests, just went with the flow.. :) But yes, integration tests seem like a better fit.

I'll convert the Tier 2 tests (finalizer, lease, cluster_working_namespace, compute_instance_working_namespace) to integration test targets and wire them into run_tests.sh. The Tier 1 tests don't need a cluster at all, so they'll stay as role-level unit tests.

collections/ansible_collections/osac/service/roles/enumerate_templates/tests/test.yml
```diff
@@ -0,0 +1,205 @@
...
+          - item.title is defined
+          - item.title | length > 0
+          - item.node_sets is defined
+          - "'fc430' in item.node_sets"
+        fail_msg: "Cluster template '{{ item.id }}' missing required fields"
+        success_msg: "Cluster template '{{ item.id }}' has correct structure"
+      loop: "{{ osac_cluster_templates }}"
+      loop_control:
+        label: "{{ item.id }}"
+      when: test_discover_all | default(false) | bool
```
akshaynadkarni, line 66: This assertion loops over ALL cluster templates and asserts each one has `fc430` in `node_sets`. Today both cluster templates (`ocp_4_17_small` and `ocp_4_17_small_github`) happen to use `fc430`, so this passes. But if anyone adds a new cluster template with a different resource class, this test will break even though the new template is perfectly valid.

Consider splitting this into two tasks: a structure validation loop that checks all templates have the required fields and a non-empty `node_sets`, and a separate assertion that checks `fc430` only on the specific templates known to use it:

```suggestion
    - name: Verify cluster template structure
      ansible.builtin.assert:
        that:
          - item.id is defined
          - item.title is defined
          - item.title | length > 0
          - item.node_sets is defined
          - item.node_sets | length > 0
        fail_msg: "Cluster template '{{ item.id }}' missing required fields"
        success_msg: "Cluster template '{{ item.id }}' has correct structure"
      loop: "{{ osac_cluster_templates }}"
      loop_control:
        label: "{{ item.id }}"
      when: test_discover_all | default(false) | bool

    - name: Verify known templates include fc430 node set
      ansible.builtin.assert:
        that:
          - "'fc430' in item.node_sets"
        fail_msg: "Template '{{ item.id }}' expected to have fc430 node set"
        success_msg: "Template '{{ item.id }}' has fc430 node set"
      loop: "{{ osac_cluster_templates | selectattr('id', 'in', ['osac.templates.ocp_4_17_small', 'osac.templates.ocp_4_17_small_github']) | list }}"
      loop_control:
        label: "{{ item.id }}"
      when: test_discover_all | default(false) | bool
```

akshaynadkarni:
> @trewest Actually there's no particular reason I put these as standalone role tests rather than integration tests, just went with the flow.. :) But yes, integration tests seem like a better fit.
> 
> I'll convert the Tier 2 tests (finalizer, lease, cluster_working_namespace, compute_instance_working_namespace) to integration test targets and wire them into run_tests.sh. The Tier 1 tests don't need a cluster at all, so they'll stay as role-level unit tests.

Nice! 
Would the role-level unit-tests run as part of CI?

akshaynadkarni (approved):
Thanks for adding all these additional tests. I used AI to review and left a suggestion.

akshaynadkarni:
@zszabo-rh I intentionally added /hold to this review. I will connect with Trey and you offline on some behavior I've observed. You can /unhold once you are ready to merge.

akshaynadkarni (approved):
Changes LGTM

PR #242, opened 2026-04-14
MGMT-23659: Use fqn instead of id for NetworkClass publishing

tzvatot:
## Summary
- Change `NetworkClassTemplate` to emit `fqn` field instead of `id`
- Update `publish_templates` role to match existing NetworkClasses by `fqn`-to-`id` mapping
- Update mock server and test playbook for new field name

## Jira
[MGMT-23659](https://redhat.atlassian.net/browse/MGMT-23659)

## Related
- Depends on fulfillment-service PR (adds `fqn` field to NetworkClass proto and generates UUID for `id`)

## Test plan
- [x] Empty scenario test passes
- [x] No-items-key edge case test passes
- [ ] Populated scenario test (pre-existing `json_query` issue in cluster templates path, unrelated)

### DIFF
--- a/collections/ansible_collections/osac/service/plugins/filter/find_template_roles.py
+++ b/collections/ansible_collections/osac/service/plugins/filter/find_template_roles.py
@@ -281,7 +281,7 @@ class NetworkClassTemplate(Base):
         return str(value)
 
     @pydantic.computed_field
-    def id(self) -> str:
+    def fqn(self) -> str:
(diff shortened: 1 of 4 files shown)


tzvatot:
Closing in favor of #243 — dropped the `fqn` field per review feedback. Now matches existing NetworkClasses by `implementation_strategy` instead, which is simpler and doesn't require a new proto field.

PR #246, opened 2026-04-15
MGMT-23769: Add default value serialization for compute instances

DakCrowder:
This PR adds the ability for other services (e.g. fulfillment-service) to fetch template default values, which allows for better validation attempts at compute instance creation time.

Additionally, a missing required field was added to `ocp_virt_vm/defaults/main.yaml` and the vendored massopencloud ocp_virt_vm default refactored a bit to place default values under the `default_spec` key.

A prerequisite to: https://github.com/osac-project/fulfillment-service/pull/414

## Summary by CodeRabbit

* **New Features**
  * Template parameters now support validation patterns.
  * Added an optional exposed_ports parameter for VM/ComputeInstance templates.
  * ComputeInstance templates now provide spec defaults for CPU, memory, boot disk, image source, and run strategy (sourced from template metadata).

* **Tests**
  * Added integration tests covering defaults merging and parameter validation.

* **Documentation**
  * Updated template development guidance and parameter-definition workflow.

* **Chores**
  * Added a lint target to the build tooling.


tzvatot (commented):
## Review: MGMT-23769 — Add default value serialization for compute instances

Clean PR that adds `spec_defaults` to compute instance templates. The `_read_yaml` refactor reduces duplication well, the Pydantic models handle camelCase→snake_case aliasing correctly, and the vendor defaults change is consistent with how `create_validate.yaml` already consumes `default_spec`. Tests cover both baseline defaults and override persistence.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 0 |
| 💡 Suggestion | 2 |

---

### 💡 `ComputeInstanceImage.source_type` — consider making optional

**File:** `collections/.../plugins/filter/find_template_roles.py:185-190`

Both `source_type` and `source_ref` are required fields on `ComputeInstanceImage`. Since the templates role's `defaults/main.yaml` only just added `sourceType` in this PR (it wasn't there before), any third-party or older role that defines `image` with only `sourceRef` will fail `model_validate()`, silently dropping the entire `spec_defaults` block (line 505 catches the exception and sets `spec_defaults = None`).

Making `source_type` optional (`str | None = pydantic.Field(default=None, ...)`) would be more resilient. `source_ref` being required when `image` is present seems reasonable.

---

### 💡 Baseline test verifies runtime fallback but not filter output

**File:** `tests/.../compute_instance_create_defaults/tasks/baseline.yml`

The test asserts on `vm_cpu_cores`, `vm_memory`, etc. — which proves `create_validate.yaml` resolved defaults correctly at runtime. But it doesn't verify that `find_template_roles.py` actually emits `spec_defaults` in the template payload. A regression in the filter (e.g., the Pydantic model failing silently) would still pass this test because the workflow falls back to `default_spec` from the role's `defaults/main.yaml` directly.

Consider adding an assertion that inspects the template discovery output to confirm `spec_defaults` is populated with expected values.

collections/ansible_collections/osac/service/plugins/filter/find_template_roles.py
```diff
@@ -303,77 +354,67 @@ class Collection(Base):
...
             path: Path to the role directory
+            subdir: Subdirectory within the role (e.g. "meta", "defaults")
+            name: Filename without extension (e.g. "main", "osac")
 
         Returns:
-            Metadata object if found and valid, None otherwise
+            Parsed YAML dict if found and valid, None otherwise
         """
-        for filename in ["osac.yaml", "osac.yml"]:
-            metadata_file: Path = path / "meta" / filename
```
adriengentil, line 316: We used to put all the information that we wanted to publish to the fulfillment-service in `meta/` folder of the ansible role. Templates are supposed to be extended by other cloud providers, so I think providing an opinionated place where to put defaults, metadata and so on for osac to pick them up will be easier to explain document later.

 Isn't it possible to use the [arg spec](https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/meta/argument_specs.yaml) to build the ComputeInstanceTemplate? And this definition would be located along the template parameter `exposed_ports`

https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_reuse_roles.html#role-argument-validation

DakCrowder: Thanks for the context on meta/ - I believe the templates _can_ be added (e.g. if they wanted to add a vm_large with more cores/disk), but would require a complete additional role.  I do think that warrants a look but not sure if it should be handled in this PR (might need a  more holistic thought process and view).

wrt the arg_spec to build the ComputeInstanceTemplate - I think the current separation is intentional but this is my first time looking through all these repos so happy to defer to you (or anybody else) on if my interpretation is incorrect.  Currently, it seems like
- spec fields are explicitly called out in the .proto, and the fulfillment service validates them applies defaults and they map (eventually) to fields in the CRD.
- template parameters are opaque to the service, and passes them through as .Any fields in the .proto.  They seem to be more flexible / just a map and template providers have a bit more flexibility and the fulfillment service doesn't really know much about them.

The spec fields seem more static / "core" fields that have different validation requirements across the stack so I think keeping them separate for now probably makes sense.

adriengentil: > fulfillment service doesn't really know much about them

`template_parameters` are published to the fulfillment-service using this `publish_template` role, and already use the arg spec to determine the default value, and the type of the field:
```
$ fulfillment-cli get computeinstancetemplate -o yaml
'@type': type.googleapis.com/osac.private.v1.ComputeInstanceTemplate
description: |
  Simple ComputeInstance
id: osac.templates.ocp_virt_vm
metadata:
  creation_timestamp: "2026-04-03T09:15:50.992192Z"
  creators:
    - system
  tenants:
    - shared
parameters:
  - default:
      '@type': type.googleapis.com/google.protobuf.StringValue
      value: 22/tcp
    description: |
      Ports to expose on the VM for ingress traffic. The syntax is a comma-separated list of `<port>/<protocol>` pairs, where `<protocol>` is either `tcp` or `udp`. For example, `22/tcp,80/tcp` will expose tcp ports 22 and 80 on the VM.
    name: exposed_ports
    type: type.googleapis.com/google.protobuf.StringValue
title: Simple ComputeInstance Template
```

That's why I think it would be the right place to put the same information about the well-known fields. The arg spec allows us to treat the Ansible Role as a function call, with all the input parameters declared in a well known place, for humans and automation.

>  I think the current separation is intentional 

These `default_arg_specs` and `default_spec` are looking like workaround because Ansible has a weird behavior where is doesn't always validate and/or set the default values depending on the context. `default_arg_specs` duplicates the information about the default value of `exposed_ports` in the arg spec.

I think I would like to see:
- A single file, in a well known place, that is the source of truth for well-known parameters and anonymous parameters where default values are specified, and their possible validations (we could imagine to have a CEL exported to the fulfillment-service when the template as stricter validations for some fields... not in the scope of this ticket)
- This pattern can be re-used for other kind of templates (cluster, network classes, ...)
- This file would be used for to compute the final value for the args for the template, and for publishing to the template to the fulfillment-service

It could be the arg spec, but it could also be also something else that we design to fit our needs, we choose the arg spec when we started the project because it was "given" by Ansible, but it's not perfect.

DakCrowder: Got it!  Really appreciate the detailed response, seeing the cli tie in is very helpful - I'll take a pass at consolidating those values in arg_spec and seeing how that shakes out

collections/ansible_collections/osac/service/plugins/filter/find_template_roles.py
```diff
@@ -180,6 +192,74 @@ class NodeSet(Base):
...
 
 
+class ComputeInstanceImage(Base):
+    """Image configuration for compute instance spec defaults."""
+
+    source_type: str = pydantic.Field(
+        ...,
+        validation_alias=pydantic.AliasChoices("sourceType", "source_type"),
+        serialization_alias="source_type",
+    )
```
akshaynadkarni, line 202: `source_type` is required here (`...`), but the CRD only supports one value today (`registry`, per the [enum in the CRD](https://github.com/osac-project/osac-operator/blob/main/config/crd/bases/osac.openshift.io_computeinstances.yaml#L118-L120)). If a future template author defines an `image` block without `source_type`, Pydantic validation fails and the `except` at line 440 silently drops the entire template metadata (not just the image field, but title, parameters, and all spec_defaults).

Since there's only one valid value, defaulting to it would prevent that silent failure:

```suggestion
    source_type: str | None = pydantic.Field(
        default="registry",
        validation_alias=pydantic.AliasChoices("sourceType", "source_type"),
        serialization_alias="source_type",
    )
```

akshaynadkarni:
@DakCrowder The `osac.yaml` consolidation looks good. One thing I noticed: the [templates collection README](https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/templates/README.md) still tells developers to define parameters in `meta/argument_specs.yaml` (e.g., the "Creating a New Cluster Template" section and the contributing guidelines). With this PR, both spec defaults and parameters now live in `meta/osac.yaml` instead.

Could you update the README to reflect the new convention? Specifically:
- Where to define `spec_defaults` (in `osac.yaml`, with the available fields and their types)
- Where to define `parameters` (in `osac.yaml`, with the `name`, `type`, `default`, `required`, `validation` schema)

Without this, a developer creating a new template would follow the current README, put their parameters in `argument_specs.yaml`, and their defaults wouldn't reach the fulfillment-service.

DakCrowder:
@akshaynadkarni Great point - updated the README.  Currently the compute instance template vs cluster are different and docs have been updated with different instructions.  I think it would be worth aligning these in a follow up pr and further fleshing out the docs between them when that happens.

collections/ansible_collections/osac/templates/roles/ocp_virt_vm/tasks/create_validate.yaml
```diff
@@ -1,21 +1,23 @@
...
-    vm_cpu_cores: "{{ (compute_instance.spec.cores | default(default_spec.cores)) | int }}"
-    vm_memory: "{{ (compute_instance.spec.memoryGiB | default(default_spec.memoryGiB)) | string + 'Gi' }}"
-    vm_boot_disk_size: "{{ (compute_instance.spec.bootDisk.sizeGiB | default(default_spec.bootDisk.sizeGiB)) | string + 'Gi' }}"
-    vm_image_source: "{{ compute_instance.spec.image.sourceRef | default(default_spec.image.sourceRef) }}"
-    vm_run_strategy: "{{ compute_instance.spec.runStrategy | default(default_spec.runStrategy) }}"
+    vm_cpu_cores: "{{ compute_instance.spec.cores | int }}"
+    vm_memory: "{{ compute_instance.spec.memoryGiB | string + 'Gi' }}"
+    vm_boot_disk_size: "{{ compute_instance.spec.bootDisk.sizeGiB | string + 'Gi' }}"
+    vm_image_source: "{{ compute_instance.spec.image.sourceRef }}"
+    vm_run_strategy: "{{ compute_instance.spec.runStrategy }}"
```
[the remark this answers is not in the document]
DakCrowder: The objective of this work was to move the spec default values into a single location (in this case osac.yaml) - adding explicit default checks with values (e.g. 2 cores) inline w/ the yaml defeats this purpose.

akshaynadkarni:
@DakCrowder 
The `compute_instance_create_defaults` and `compute_instance_create_validation` tests are offline (no K8s API calls, just Ansible logic against a YAML fixture). But they're listed in the `WORKFLOWS` array in `run_tests.sh`, which runs after `setup_test_env.sh` creates a kind cluster.

[PR #239](https://github.com/osac-project/osac-aap/pull/239) introduced a convention for role-level tests: they live inside the role at `roles/<name>/tests/test.yml` (e.g., [`enumerate_templates/tests/test.yml`](https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/service/roles/enumerate_templates/tests/test.yml)) with scenario guards (`when: test_var | default(false) | bool`), and are listed under `ROLE_TESTS` or `ROLE_SCENARIO_TESTS` in `run_tests.sh`. Since these tests exercise the `ocp_virt_vm` role's `create_validate.yaml` logic, they'd fit at `collections/ansible_collections/osac/templates/roles/ocp_virt_vm/tests/test.yml`.

Not blocking, but aligning with the existing convention would keep the test structure consistent and let these tests run without a cluster. 

@zszabo-rh Can you please confirm you are in agreement with what I'm saying above :point_up: ? I want to make sure we maintain the established pattern.

akshaynadkarni (approved):
Changes LGTM. 
I have posted a question for Zoltan here: https://github.com/osac-project/osac-aap/pull/246#issuecomment-4338251795

Depending on his response, you can merge or update the code as part of this PR.

zszabo-rh:
@DakCrowder @akshaynadkarni 
> @zszabo-rh Can you please confirm you are in agreement with what I'm saying above ☝️ ? I want to make sure we maintain the established pattern.

Yes, I'd confirm Akshay's suggestion, the tests should live at `roles/ocp_virt_vm/tests/test.yml` with per-scenario guards.

[PR #239](https://github.com/osac-project/osac-aap/pull/239) established two tiers:
- Role-level tests (`roles/<name>/tests/test.yml`) for offline tests that don't need K8s
- Integration test targets (`tests/integration/targets/`) for tests that need the kind cluster

Dakota's tests match Tier 1 pattern. Putting them in `tests/integration/targets/` means they unnecessarily depend on the kind cluster setup, and they're testing a specific role's behavior, not a cross-component workflow.

akshaynadkarni:
Thanks @zszabo-rh . 

@DakCrowder See Zoltan's comment [above](https://github.com/osac-project/osac-aap/pull/246#issuecomment-4341135338). Will you be addressing the move as part of a separate PR?

DakCrowder:
Thanks for the reply @zszabo-rh!!

@akshaynadkarni Yes I'll get a PR up to get those sorted

PR #251, opened 2026-04-20
NO_ISSUE: fix vendored aap_configuration regex_replace corrupting omit sentinel

omer-vishlitzky:
Backport of the upstream fix from `infra.aap_configuration` (fixed between v3.4.1 and v4.x).

## Problem

When `variables`, `source_vars`, `injectors`, or `messages` are not defined in the
inventory/host/credential_type/notification config, the vendored v3.4.1 code pipes
the Ansible `omit` sentinel through `regex_replace`:

```yaml
variables: "{{ ... | default(omit) | regex_replace(...) }}"
```

This corrupts the `omit` sentinel, causing the `ansible.controller` module to receive
it as a literal string. Since the module declares `variables=dict(type='dict')`, it
fails with:

```
argument 'variables' is of type str and we were unable to convert to dict
```

This breaks the AAP bootstrap job — all 5 controller inventories fail to create.

## Fix

Apply the upstream pattern: only call `regex_replace` when the variable is actually
defined, keeping `omit` on a separate code path:

```yaml
variables: "{{ ... | regex_replace(...) if (var is defined and var) else (... omit) }}"
```

## Affected roles (all 6 fixed)

- `controller_inventories` — `variables`
- `controller_inventory_sources` — `source_vars`
- `controller_credential_types` — `injectors`
- `controller_host_groups` — `variables`
- `controller_hosts` — `variables`
- `controller_notification_templates` — `messages`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty or undefined configuration variables in deployments: placeholder-cleanup transformations are now skipped when values are missing, preventing unnecessary processing and reducing unexpected side-effects.

* **Chores**
  * Updated approvers and reviewers list to include an additional team member.

### DIFF
--- a/vendor/ansible_collections/infra/aap_configuration/roles/controller_credential_types/tasks/main.yml
+++ b/vendor/ansible_collections/infra/aap_configuration/roles/controller_credential_types/tasks/main.yml
@@ -4,7 +4,7 @@
     name: "{{ __controller_credential_type_item.name | mandatory }}"
     new_name: "{{ __controller_credential_type_item.new_name | default(omit, true) }}"
     description: "{{ __controller_credential_type_item.description | default(('' if controller_configuration_credential_types_enforce_defaults else omit), true) }}"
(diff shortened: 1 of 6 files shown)


akshaynadkarni (approved):
LGTM

PR #254, opened 2026-04-21
MGMT-24061: disable scm_update_on_launch to speed up job processing

adriengentil:
## Summary

- `scm_update_on_launch: true` on the AAP project caused an SCM sync before every job execution, adding unnecessary overhead to every provisioning/deprovisioning operation
- Set `scm_update_on_launch: false` — SCM updates are already handled by the periodic `{{ aap_prefix }}-sync` schedule (every 10 minutes)

## Test plan

- [x] Config-as-code applies cleanly with updated project definition
- [x] AAP jobs launch without triggering an SCM sync

Fixes: https://redhat.atlassian.net/browse/MGMT-24061

Co-Authored-By: Claude Sonnet 4.6 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **Chores**
  * Updated project configuration: SCM repository updates will no longer trigger automatically when projects launch.


akshaynadkarni (approved):
LGTM

PR #255, opened 2026-04-21
Fix empty template lists caused by non-blocking stdin in AAP

trewest:
When running inside AAP, stdin is set to non-blocking mode. The ansible-galaxy subprocess in find_template_roles inherits this non-blocking stdin, causing it to exit with an error. The collection query failure is caught and skipped, resulting in empty template lists.

Pass stdin=subprocess.DEVNULL so ansible-galaxy gets a clean blocking file descriptor instead of inheriting the parent's non-blocking one.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue with the service collection filter that could cause unexpected behavior when processing template role information by preventing the ansible-galaxy command from attempting to read from standard input.


trewest:
Handled by https://github.com/osac-project/osac-aap/pull/256

PR #257, opened 2026-04-21
NO_ISSUE: fix restart check failing on AnsibleTaggedDateTime

omer-vishlitzky:
## Summary
- Add `| string` before `| length` in the restart check in `ocp_virt_vm/tasks/create_resources.yaml`

## Problem
ansible-core 2.20 returns `_AnsibleTaggedDateTime` for datetime fields like `restartRequestedAt`. The `length` filter calls `len()` on this object, which fails:

```
The filter plugin 'ansible.builtin.length' failed: object of type '_AnsibleTaggedDateTime' has no len()
```

This causes every restart AAP job to fail, leaving `lastRestartedAt` empty.

## Fix
Convert to string before checking length: `| default('') | string | length > 0`

### DIFF
--- a/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/tasks/create_resources.yaml
+++ b/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/tasks/create_resources.yaml
@@ -109,12 +109,12 @@
   ansible.builtin.set_fact:
     restart_requested: >-
       {{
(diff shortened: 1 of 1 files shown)


akshaynadkarni (approved):
Interesting. I wonder if `datetime` has some other functionality that allows you to do the same things vs. having to convert to a string. 

Anyways, for now this LGTM.

PR #245, opened 2026-04-15
MGMT-23607: Add CLAUDE.md for AI-assisted development

eranco74:
Add comprehensive CLAUDE.md documentation to guide AI coding assistants (Claude Code, Copilot, etc.) when working with the osac-aap codebase.

## Summary

- Comprehensive documentation for AI assistants and developers
- Covers repository structure, collections, playbooks, and template roles
- Documents development workflows (uv setup, vendoring, testing)
- Explains AAP integration patterns and Config-as-Code
- Details networking implementation (cudn_net role)
- Lists common pitfalls and cross-repo coordination patterns

## Key Sections

- **Repository Structure**: Collections organization (osac.service, osac.templates, osac.workflows)
- **Playbook Architecture**: Standard pattern for AAP job templates
- **Template Roles**: How to create new infrastructure implementations with `meta/osac.yaml`
- **Networking (cudn_net)**: CUDN provisioning for VirtualNetwork/Subnet/SecurityGroup
- **Service Roles**: Common utilities (common, finalizer, wait_for, etc.)
- **Testing**: ansible-lint, ansible-test integration tests
- **Common Pitfalls**: implementation_strategy naming, namespace labels, vendoring issues

## Benefits

- Improves developer onboarding
- Provides context for AI assistants to generate accurate Ansible code
- Documents conventions and patterns across the codebase
- Clarifies osac-templates merge into osac-aap

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guidance for Ansible authoring, repository layout, role/template conventions, and playbook patterns.
  * Documented networking rules for CUDN, namespace label requirements, and how NetworkPolicy is derived from security groups.
  * Included remote-cluster provisioning steps, lint/test/setup/re-vendor commands, common pitfalls with fixes, and a PR/dev checklist.

PR #268, opened 2026-04-28
MGMT-24178, MGMT-24196: Netris CaaS fixes / improvements

danmanor:
## Summary

- Make the Netris CaaS create flow fully idempotent so it supports scale up/down without destroying existing infrastructure
- Server clusters are updated in place (PUT) instead of deleted and recreated, preserving the VPC
- NAT rules (SNAT/DNAT) are looked up first; if they exist, their IPs are reused and IPAM allocation is skipped
- NMState config is only applied via SSH to newly added agents, preventing failures on already-provisioned servers with OVS bridges
- Add a "wait for nodes ready" step before cluster operators check
- Add request timeouts to all Netris API and k8s polling calls

## Summary by CodeRabbit

* **New Features**
  * Node readiness wait step with configurable retries, delay, expected count and timeouts.
  * Conditional NAT/DNAT handling to detect and reuse existing rules and IPs.
  * Server cluster flow now updates existing clusters or creates when missing.

* **Bug Fixes**
  * Preserve existing infrastructure and avoid destructive deletions; improved idempotency.
  * Apply NMState only to newly added agents when appropriate.

* **Chores**
  * Test environment scripts switched to Podman-backed provider.
  * .gitignore extended to ignore additional local tooling directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -176,3 +176,4 @@ tests/integration/kubeconfig-*
 
 # CLAUDE.md
 CLAUDE.md
+.claude/
--- a/collections/ansible_collections/netris/controller/roles/nat/tasks/create.yaml
+++ b/collections/ansible_collections/netris/controller/roles/nat/tasks/create.yaml
@@ -42,9 +42,30 @@
       }) }}
   when: nat_action == 'snat'
 
-- name: Reset existing NAT check (avoid stale state from prior delete call)
+- name: Get existing NAT rules
+  ansible.builtin.uri:
+    url: "{{ netris_controller_url }}/api/v2/nat"
+    method: GET
+    headers:
+      Cookie: "connect.sid={{ netris_session_cookie }}"
+    return_content: true
+    status_code: 200
+    timeout: "{{ netris_timeout | default(30) }}"
+    validate_certs: "{{ netris_validate_certs | default(true) }}"
+  register: _nat_list_resp
+  no_log: true
+
+- name: Find NAT rule by name
   ansible.builtin.set_fact:
-    _existing_nat: null
+    _existing_nat: "{{ (_nat_list_resp.json.data | default([])) | selectattr('name', 'equalto', nat_name) | list | first | default(none) }}"
+  when: _nat_list_resp.json is mapping
+
+- name: Set nat facts from existing rule (skip creation)
+  ansible.builtin.set_fact:
+    nat_id: "{{ _existing_nat.id }}"
+    nat_already_existed: true
+    nat_existing_rule: "{{ _existing_nat }}"
+  when: _existing_nat is defined and _existing_nat is not none
 
 - name: Create NAT rule
   ansible.builtin.uri:
@@ -57,13 +78,17 @@
     body: "{{ _nat_body }}"
     return_content: true
     status_code: [200, 201]
+    timeout: "{{ netris_timeout | default(30) }}"
     validate_certs: "{{ netris_validate_certs | default(true) }}"
   register: _nat_create_resp
+  when: _existing_nat is not defined or _existing_nat is none
 
-- name: Set nat_id from create response
+- name: Set nat facts from create response
   ansible.builtin.set_fact:
     nat_id: "{{ _nat_create_resp.json.data.id | default(_nat_create_resp.json.data) }}"
+    nat_already_existed: false
   when:
-    - not (_nat_create_resp.skipped | default(false))
+    - _existing_nat is not defined or _existing_nat is none
+    - _nat_create_resp is not skipped
     - _nat_create_resp is changed or (_nat_create_resp.status | default(0)) in [200, 201]
     - _nat_create_resp.json is mapping
--- a/collections/ansible_collections/netris/controller/roles/server_cluster/tasks/create.yaml
+++ b/collections/ansible_collections/netris/controller/roles/server_cluster/tasks/create.yaml
@@ -4,6 +4,24 @@
     name: netris.controller.auth
   when: netris_session_cookie is not defined
 
+- name: Get existing server clusters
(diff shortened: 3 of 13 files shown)


tzvatot (commented):
## Review: MGMT-24178, MGMT-24196: Netris CaaS fixes / improvements

Solid idempotency improvement — the core design of "lookup existing resource before create, reuse if found" is the right pattern for making the CaaS create flow re-runnable. The NAT reuse, server cluster PUT-update, and NMState scoping to new agents are all well-motivated. The `wait_for_nodes` addition fills a real gap in the install sequence.

| Category | Count |
|----------|-------|
| 🔴 Critical | 1 |
| 🟡 Important | 3 |
| 💡 Suggestion | 2 |

---

### 🔴 DNAT IP allocation logic breaks when exactly one of two rules exists

**File:** `netris/steps/roles/external_access/tasks/create.yaml:79-117`

The IPAM allocation count is computed as the sum of missing DNAT rules (0, 1, or 2). But the IP assignment logic has four branches based on `_missing_api_dnat` and `_missing_ingress_dnat` combinations, and two of them assign `ipam_allocated_ips[0]` — which is correct when only one IP is allocated. However, the "both missing" branch assigns `[0]` and `[1]`, and the "only one missing" branches assign `[0]`. This works.

The actual bug: when `_dnat_ipam_count` is `0` (both rules exist), the IPAM role is skipped, but `ipam_allocated_ips` is still referenced in the conditional branches. If Ansible evaluates those `set_fact` tasks even with `when: false`, the variable reference could fail on some Ansible versions. More critically: **the `when` conditions on the "Set API DNAT IP from IPAM" tasks use `_missing_api_dnat | bool` and `_missing_ingress_dnat | bool`, but these are set as strings** (`"True"` / `"False"` from Jinja), and `"False" | bool` is `True` in Ansible because non-empty strings are truthy.

**Recommendation:** Use native booleans, not string booleans:

```yaml
- name: Count missing DNAT IPs
  ansible.builtin.set_fact:
    _missing_api_dnat: >-
      {{ _existing_api_dnat is not defined or _existing_api_dnat is none }}
```

This produces the string `"True"` or `"False"`, and `"False" | bool` evaluates to `false` in Ansible — actually this works because Ansible's `bool` filter handles the strings "True"/"False" correctly. But it's fragile. Test with `ansible -m debug -a "msg={{ 'False' | bool }}"` to confirm on your Ansible version.

Better: set them as actual booleans using `| bool` at assignment time:

```yaml
    _missing_api_dnat: "{{ (_existing_api_dnat is not defined or _existing_api_dnat is none) | bool }}"
```

---

### 🟡 `netris_nat_snat_ip` inconsistent format — existing rule has no `/32` suffix

**File:** `netris/steps/roles/cluster_infra/tasks/create.yaml:184-185`

When reusing an existing SNAT rule:
```yaml
    netris_nat_snat_ip: "{{ _existing_snat.snatToIP }}"
```

When allocating a new IP:
```yaml
    netris_nat_snat_ip: "{{ ipam_allocated_ips[0] }}/32"
```

The new-allocation path appends `/32` but the reuse path takes the raw value from the API, which may or may not include the mask. The downstream consumer (`nat_snat_to_ip`) receives inconsistent formats. CodeRabbit flagged this too.

**Recommendation:** Normalize on the reuse path:
```yaml
    netris_nat_snat_ip: "{{ _existing_snat.snatToIP | ansible.utils.ipaddr('host/prefix') }}"
```

Or strip `/32` from the new-allocation path if the API doesn't return it.

---

### 🟡 Server cluster update body only sends `servers` and `tags` — may reset other fields

**File:** `netris/controller/roles/server_cluster/tasks/create.yaml:63-67`

```yaml
    _update_body:
      servers: "{{ _server_cluster_servers | default([]) }}"
      tags: "{{ server_cluster_tags | default([]) }}"
```

The PUT request sends only `servers` and `tags`. If the Netris API treats PUT as a full replacement (not a merge), other fields like `template`, `admin`, and `vpc` could be reset to defaults. This would silently break the cluster.

**Recommendation:** Either:
1. Confirm that the Netris API does a merge-style PUT (only updates provided fields), or
2. Include all fields from `_existing_server_cluster` in the update body, overriding only `servers` and `tags`:
```yaml
    _update_body: "{{ _existing_server_cluster | combine({'servers': _server_cluster_servers, 'tags': server_cluster_tags}) }}"
```

---

### 🟡 `no_log: true` missing on server cluster update task

**File:** `netris/controller/roles/server_cluster/tasks/create.yaml:74-84`

The update task sends a request with `connect.sid` cookie but doesn't have `no_log: true`, while the create task (line 121) and all NAT tasks do. Auth cookies will appear in Ansible logs and CI output.

**Recommendation:** Add `no_log: true` to the update task.

---

### 💡 NMState split into "create CRs for all" + "apply live on new only" — intent is clear but naming could be clearer

**File:** `netris/steps/roles/cluster_infra/tasks/create.yaml:232-249`

The first call creates NMStateConfig CRs with `nmstate_config_apply_live: false`, the second applies live changes only to `_newly_added_agents`. The role must support this `apply_live` flag — worth verifying it exists and is respected.

---

### 💡 `.claude/` added to `.gitignore` — may conflict with PR #245

**File:** `.gitignore:179`

PR #245 adds `CLAUDE.md` and `.claude/rules/` to the repo. This PR adds `.claude/` to `.gitignore`, which would make those files invisible to git. If #245 merges first, this line would cause the rules files to be ignored. If this PR merges first, #245 would need to add exceptions.

**Recommendation:** Coordinate with PR #245. If `.claude/rules/` should be tracked, use:
```
.claude/*
!.claude/rules/
```

tzvatot (approved):
/lgtm

PR #274, opened 2026-04-30
MGMT-23997: Remove promoted fields from template_parameters

tzvatot:
## Summary

Remove `pull_secret` and `ssh_public_key` from `argument_specs.yaml` in cluster templates. These are now explicit ClusterOrder spec fields (MGMT-23939, MGMT-23938, [MGMT-23941](https://redhat.atlassian.net/browse/MGMT-23941)).

## Why

The `cluster_settings` role already reads these from the CR's explicit spec fields first, falling back to `template_parameters`. But the fulfillment-service validates template parameters against `argument_specs.yaml` at create time — since `pull_secret` was marked `required: true`, users couldn't use `--pull-secret-file` (explicit field) without also passing it as a template parameter.

## Changes

- **`ocp_4_17_small/meta/argument_specs.yaml`** — removed `pull_secret` (required) and `ssh_public_key` from `template_parameters.options`
- **`ocp_4_17_small_github/meta/argument_specs.yaml`** — same (GitHub-specific params like `github_client_id` kept)
- Vendored copies updated to match

## Test plan

- [ ] Re-publish templates on vmaas-dev after merge
- [ ] Verify `fulfillment-cli create cluster --pull-secret-file <file> --ssh-public-key-file <file>` works without `--template-parameter-file pull_secret=...`

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Removed template validation for pull_secret and ssh_public_key across OCP 4.17 small role templates, making those parameters no longer enforced at template-validation time.
* **Bug Fixes**
  * Added enforcement that a pull secret must be resolved from all configured sources; deployments now fail with a clear message and guidance if no pull secret is found.


trewest (changes requested):
From Claude:

  Suggestion: Add early validation for empty pull_secret in cluster_settings                                                                                                                                                                  
                                                                                                                                                                                                                                              
  With pull_secret removed from argument_specs.yaml, the fulfillment-service no longer rejects requests that omit it. The cluster_settings role has a three-tier fallback (explicit CR field → templateParameters → provider default Secret),
  but if all three miss, cluster_settings_pull_secret silently resolves to an empty string. The HostedCluster then fails with a cryptic image-pull error that gives no indication the pull secret is missing.

  Consider adding a fail task in cluster_settings/tasks/main.yml after the provider default Secret fallback (after line 68), before the merge step:

  - name: Fail if pull_secret could not be resolved
    ansible.builtin.fail:
      msg: >-
        pull_secret is empty after checking all sources
        (spec.pullSecret, templateParameters, and the provider default Secret
        '{{ cluster_settings_default_credentials_secret_name }}'
        in namespace '{{ cluster_settings_default_credentials_secret_namespace }}').
        Supply a pull secret via --pull-secret-file, spec.pullSecret, or configure
        the provider default Secret.
    when: not cluster_settings_pull_secret

  This follows the existing validation pattern used in tenant_storage_class, cluster_working_namespace, and nmstate_config roles. ssh_public_key doesn't need this since it's already required: false in the hosted_cluster argument spec and
  clusters work without it.

adriengentil:
I think we can just remove the arg spec because it's almost empty now :thinking:

tzvatot:
> I think we can just remove the arg spec because it's almost empty now 🤔

Thanks - I removed it.

tzvatot:
> From Claude:
> 
> Suggestion: Add early validation for empty pull_secret in cluster_settings
>  ...

Added, thanks!

PR #276, opened 2026-04-30
MGMT-24232: add ci.steps network backend and ocp_ci_small template

omer-vishlitzky:
https://issues.redhat.com/browse/MGMT-24232

Adds a `ci.steps` network steps collection for CI environments without
ESI or Netris. Same pattern as `massopencloud.steps` and `netris.steps`,
plugged in via `NETWORK_STEPS_COLLECTION=ci.steps`.

**ci.steps.cluster_infra**: selects, labels, and approves agents without
ESI network attachment. Creates per-cluster CAPI RBAC for HyperShift
agent access.

**ci.steps.external_access**: noop (no floating IPs or DNS in CI).

**ocp_ci_small template**: identical install flow to ocp_4_17_small,
different defaults (1 agent of class ci-worker).


collections/ansible_collections/osac/service/roles/hosted_cluster/tasks/create_hosted_cluster.yaml
```diff
@@ -44,9 +44,9 @@
           name: "sshkey-cluster-{{ hosted_cluster_name }}"
         networking:
           clusterNetwork:
-            - cidr: "{{ hosted_cluster_pod_cidr }}"
+            - cidr: 10.132.0.0/14
           serviceNetwork:
-            - cidr: "{{ hosted_cluster_service_cidr }}"
+            - cidr: 172.31.0.0/16
```
akshaynadkarni: Why are we hardcoding IPs?

collections/ansible_collections/osac/templates/roles/ocp_ci_small/meta/argument_specs.yaml
akshaynadkarni, line 1: @DakCrowder Can you verify whether/not this structure aligns with the new format of defining templates and their defaults in our repos? 

`ocp_ci_small` is a new template being created. I want to make sure the values flow correctly to the `fulfillment-service`. 

Linking the README here: 
https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/templates/README.md

akshaynadkarni: @DakCrowder This is for CaaS and not VMaaS. We haven't touched the default translation logic in CaaS. So, I think we are good with this.

akshaynadkarni (approved):
LGTM

collections/ansible_collections/osac/templates/roles/ocp_ci_small/meta/argument_specs.yaml
DakCrowder: It currently _doesn't_ fully align - however some of the changes for the cluster templates were only merged today:  https://github.com/osac-project/fulfillment-service/pull/474.

There is now the `spec_defaults` field in the cluster template, and similarly to the vm we can now look to define those in osac.yaml if there are any values we want to pre-populate.  We can also move the `template_parameters` into the osac.yaml under the `parameters` key however both would require some role changes to where they are read from.

Example osac.yaml from the vm template that defines the `spec_defaults` as well as `parameters` that eventually make their way into fulfillment-service:
https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/meta/osac.yaml#L9

PR #279, opened 2026-05-03
MGMT-24176: add BMH-based agent import playbook with AAP config-as-code

danmanor:
Single-file playbook that imports bare metal servers as Assisted Installer agents using BareMetalHost CRs via Ironic. Reads server inventory from a mounted ConfigMap, reconciles BMH CRs (create missing, delete stale), waits for Agent registration, and labels agents with resource class and Netris server name metadata.

Config-as-code changes add a job template, a 10-minute periodic schedule, and mount the import-agents-inventory ConfigMap into the cluster-fulfillment-ig pod spec.

Includes a sample inventory file for local testing:
```
  ansible-playbook playbook_osac_import_agents.yml \
    -e import_agents_inventory_path=samples/import_agents_servers.yml
```

## Summary by CodeRabbit

* **New Features**
  * Bare metal server import feature to automatically provision Agent resources from inventory.
  * Scheduled import job in AAP running every 10 minutes.
  * Automatic reconciliation managing resource lifecycle (creation and cleanup).

* **Documentation**
  * New guide for bare metal server import workflow, setup, and operations.

* **Chores**
  * Sample configuration files for BMC credentials and import parameters.

### DIFF
--- a/playbook_osac_import_agents.yml
+++ b/playbook_osac_import_agents.yml
@@ -1,0 +1,205 @@
+---
+- name: Import bare metal servers as Assisted Installer agents
+  hosts: localhost
+  gather_facts: false
+
+  vars:
+    import_agents_namespace: "{{ default_agent_namespace }}"
+    import_agents_infraenv_name: "{{ hosted_cluster_default_infraenv }}"
+    import_agents_inventory_path: "/var/config/import-agents/servers.yml"
+    import_agents_pull_secret_name: "pull-secret"
+    import_agents_managed_by_label: "osac.openshift.io/managed-by=import-agents"
+    import_agents_disable_bmc_cert_verification: true
+
+  tasks:
+    - name: Verify inventory file exists
+      ansible.builtin.stat:
+        path: "{{ import_agents_inventory_path }}"
+      register: _inventory_file
+
+    - name: Fail if inventory file is missing
+      ansible.builtin.fail:
+        msg: >-
+          Inventory file not found at {{ import_agents_inventory_path }}.
+          Ensure the import-agents-inventory ConfigMap is created and mounted.
+      when: not _inventory_file.stat.exists
+
+    - name: Load server inventory from file
+      ansible.builtin.include_vars:
+        file: "{{ import_agents_inventory_path }}"
+        name: import_agents_inventory
+
+    - name: Set server list
+      ansible.builtin.set_fact:
+        import_agents_servers: "{{ import_agents_inventory.servers }}"
+
+    - name: Display servers to import
+      ansible.builtin.debug:
+        var: import_agents_servers
+
+    - name: Ensure InfraEnv exists
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: agent-install.openshift.io/v1beta1
+          kind: InfraEnv
+          metadata:
+            name: "{{ import_agents_infraenv_name }}"
+            namespace: "{{ import_agents_namespace }}"
+          spec:
+            pullSecretRef:
+              name: "{{ import_agents_pull_secret_name }}"
+
+    - name: Get existing managed BareMetalHost CRs
+      kubernetes.core.k8s_info:
+        api_version: metal3.io/v1alpha1
+        kind: BareMetalHost
+        namespace: "{{ import_agents_namespace }}"
+        label_selectors:
+          - "{{ import_agents_managed_by_label }}"
+      register: import_agents_existing_bmh
+
+    - name: Extract existing BMH names
+      ansible.builtin.set_fact:
+        import_agents_existing_bmh_names: >-
+          {{ import_agents_existing_bmh.resources | map(attribute='metadata.name') | list }}
+
+    - name: Display existing managed BMHs
+      ansible.builtin.debug:
+        var: import_agents_existing_bmh_names
+
+    - name: Reconcile BareMetalHost for each server
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: metal3.io/v1alpha1
+          kind: BareMetalHost
+          metadata:
+            name: "{{ server.name }}"
+            namespace: "{{ import_agents_namespace }}"
+            labels:
+              osac.openshift.io/managed-by: import-agents
+              infraenvs.agent-install.openshift.io: "{{ import_agents_infraenv_name }}"
+            annotations:
+              inspect.metal3.io: disabled
+              bmac.agent-install.openshift.io/hostname: "{{ server.name }}"
+          spec:
+            automatedCleaningMode: disabled
+            bmc:
+              address: "{{ server.bmc_url }}"
+              credentialsName: "{{ server.bmc_secret }}"
+              disableCertificateVerification: >-
+                {{ server.disable_bmc_cert_verification
+                   | default(import_agents_disable_bmc_cert_verification) }}
+            bootMACAddress: "{{ server.boot_mac }}"
+            online: true
+      loop: "{{ import_agents_servers }}"
+      loop_control:
+        loop_var: server
+        label: "{{ server.name }}"
+
+    - name: Compute stale BMH names
+      ansible.builtin.set_fact:
+        import_agents_stale_bmh_names: >-
+          {{
+            import_agents_existing_bmh_names |
+            difference(import_agents_servers | map(attribute='name') | list)
+          }}
+
+    - name: Display stale BMHs to delete
+      ansible.builtin.debug:
+        var: import_agents_stale_bmh_names
+
+    - name: Get agents linked to stale BMHs
+      kubernetes.core.k8s_info:
+        api_version: agent-install.openshift.io/v1beta1
+        kind: Agent
+        namespace: "{{ import_agents_namespace }}"
+        label_selectors:
+          - "agent-install.openshift.io/bmh={{ bmh_name }}"
+      register: import_agents_stale_agents
+      loop: "{{ import_agents_stale_bmh_names }}"
+      loop_control:
+        loop_var: bmh_name
+        label: "{{ bmh_name }}"
+
+    - name: Delete stale Agent CRs
+      kubernetes.core.k8s:
+        state: absent
+        api_version: agent-install.openshift.io/v1beta1
+        kind: Agent
+        namespace: "{{ import_agents_namespace }}"
+        name: "{{ agent.metadata.name }}"
+      loop: >-
+        {{
+          import_agents_stale_agents.results |
+          map(attribute='resources') |
+          flatten
+        }}
+      loop_control:
+        loop_var: agent
+        label: "{{ agent.metadata.name }}"
+
+    - name: Delete stale BareMetalHost CRs
+      kubernetes.core.k8s:
+        state: absent
+        api_version: metal3.io/v1alpha1
+        kind: BareMetalHost
+        namespace: "{{ import_agents_namespace }}"
+        name: "{{ bmh_name }}"
+      loop: "{{ import_agents_stale_bmh_names }}"
+      loop_control:
+        loop_var: bmh_name
+        label: "{{ bmh_name }}"
+
+    - name: Wait for each server to register as an Agent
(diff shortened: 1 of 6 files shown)


tzvatot (commented):
## PR #279: MGMT-24176: add BMH-based agent import playbook with AAP config-as-code

This PR adds a new single-file playbook that reconciles a file-based server inventory against BareMetalHost and Agent CRs — creating missing BMHs, deleting stale ones, waiting for agent registration, and labeling agents. It includes AAP config-as-code for a job template with a 10-minute schedule, documentation, and sample files. The overall structure is sound and the reconciliation logic is well thought out (delete-agent-before-BMH ordering, managed-by label scoping, MAC-based agent matching).

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 3 |
| 💡 Suggestion | 2 |

---

### 🟡 BMH updates silently skipped for existing servers

**File:** `playbook_osac_import_agents.yml:67-90`

The `when: server.name not in import_agents_existing_bmh_names` guard skips the `kubernetes.core.k8s` task entirely for any server whose BMH already exists. This means if an operator changes a server's `bmc_url`, `bmc_secret`, or `boot_mac` in the inventory, the update will never be applied — the playbook will silently skip it.

Since `kubernetes.core.k8s` with `state: present` already does an apply (create-or-patch), removing the `when` clause would make the task idempotent for both creation and updates. The only cost is extra API calls on each run, but these are lightweight patch operations.

**Recommendation:** Remove the `when` guard and let the k8s module handle idempotency:

```yaml
    - name: Reconcile BareMetalHost for each server
      kubernetes.core.k8s:
        state: present
        definition:
          # ... same definition ...
      loop: "{{ import_agents_servers }}"
      loop_control:
        loop_var: server
        label: "{{ server.name }}"
```

If the extra API calls are a concern (many servers on a 10-minute schedule), consider a `changed_when` annotation or a separate "diff" task — but don't silently skip updates.

---

### 🟡 Job template missing `allow_simultaneous` and `ask_variables_on_launch`

**File:** `collections/ansible_collections/osac/config_as_code/roles/aap/vars/controller.yml:225-234`

Every other job template in this file sets `allow_simultaneous: true` and `ask_variables_on_launch: true`. The new `import-agents` template omits both.

For `allow_simultaneous`: omitting it is arguably correct for a reconciliation playbook (you don't want concurrent runs contending on the same BMH/Agent CRs). But this relies on the AAP default rather than being explicit, and breaks the pattern established by every other template. If it's intentional, make it explicit.

For `ask_variables_on_launch`: without this, operators can't override variables (e.g., `import_agents_namespace`, `import_agents_inventory_path`) when manually triggering the job template. This is particularly useful for debugging or testing against a different namespace.

**Recommendation:**

```yaml
  - name: "{{ aap_prefix }}-import-agents"
    # ...existing fields...
    allow_simultaneous: false
    ask_variables_on_launch: true
    verbosity: 0
```

---

### 🟡 Local testing command in docs will fail — missing required variables

**File:** `docs/import-agents.md:160-162` and `samples/import_agents_extra_vars.yml`

The documented local testing command is:

```bash
ansible-playbook playbook_osac_import_agents.yml \
  -e import_agents_inventory_path=samples/import_agents_extra_vars.yml
```

But the playbook vars reference `default_agent_namespace` and `hosted_cluster_default_infraenv`, which are defined in `group_vars/all/infrastructure.yaml`. When running locally without the AAP inventory, these are undefined and the playbook will fail with an "undefined variable" error.

**Recommendation:** Either add these to the sample extra vars file:

```yaml
# samples/import_agents_extra_vars.yml
default_agent_namespace: "hardware-inventory"
hosted_cluster_default_infraenv: "infraenv"
servers:
  - name: ostest-extraworker-0
    # ...
```

Or use literal defaults directly in the playbook vars (like other playbooks in this repo do):

```yaml
vars:
    import_agents_namespace: "hardware-inventory"
    import_agents_infraenv_name: "infraenv"
```

---

### 💡 ConfigMap mounted at `/var/secrets/` path

**File:** `collections/ansible_collections/osac/config_as_code/roles/aap/vars/controller.yml:455-461`

The `import-agents-inventory` ConfigMap is mounted at `/var/secrets/import-agents/`. Using `/var/secrets/` for a ConfigMap is misleading — it suggests sensitive data when the content is a server inventory (names, MACs, resource classes). Consider `/var/config/import-agents/` to match the data source type.

---

### 💡 Schedule frequency vs. wait timeout

**File:** `collections/ansible_collections/osac/config_as_code/roles/aap/vars/controller.yml:402` and `playbook_osac_import_agents.yml:143-155`

The schedule runs every 10 minutes, but the wait task allows up to 15 minutes per server (90 retries * 10s delay). With multiple new servers, a single playbook run can exceed 10 minutes, potentially queuing scheduled runs. This is handled correctly because `allow_simultaneous` defaults to false, so AAP will skip/queue. But worth documenting this behavior — operators should know that adding many servers at once will naturally take longer than the schedule interval.

---

**Overall:** Well-structured playbook with good reconciliation logic and thorough documentation. The three important items are straightforward to address. Approve with requested changes.

eranco74 (commented):
Nice work — the reconciliation logic is solid. tzvatot already covered the main items so I'm only adding what wasn't mentioned yet.

playbook_osac_import_agents.yml
```diff
@@ -0,0 +1,189 @@
...
+        kind: Agent
+        namespace: "{{ import_agents_namespace }}"
+      register: import_agents_final_agents
+
+    - name: Configure agent metadata
+      kubernetes.core.k8s:
+        state: present
+        definition:
+          apiVersion: agent-install.openshift.io/v1beta1
+          kind: Agent
```
eranco74, line 183: There's a subtle race condition here. The wait task (line 143) confirms the agent exists, but then a *separate* `k8s_info` call fetches the agent list at line 161. If an agent disappears between those two calls, `mac_to_agent_name` returns `None` and this task would try to patch an Agent named `"None"`.

Consider either:
- Reusing `import_agents_current_agents` from the wait task instead of fetching again, or
- Adding `failed_when` to guard against a `None` name:
```yaml
      register: import_agents_label_result
      failed_when: >
        [server.boot_mac] |
        osac.service.mac_to_agent_name(import_agents_final_agents.resources) is none
```

playbook_osac_import_agents.yml
```diff
@@ -0,0 +1,189 @@
...
+            bmc:
+              address: "{{ server.bmc_url }}"
+              credentialsName: "{{ server.bmc_secret }}"
+              disableCertificateVerification: true
+            bootMACAddress: "{{ server.boot_mac }}"
+            online: true
+      loop: "{{ import_agents_servers }}"
+      loop_control:
+        loop_var: server
+        label: "{{ server.name }}"
```
eranco74, line 99: `disableCertificateVerification: true` is hardcoded for every BMH. Fine for dev/lab but worth making configurable — either per-server in the inventory format (`disable_cert_verification: true`) or as a playbook-level variable with an explicit default.

playbook_osac_import_agents.yml
```diff
@@ -0,0 +1,189 @@
...
+  vars:
+    import_agents_namespace: "{{ default_agent_namespace }}"
+    import_agents_infraenv_name: "{{ hosted_cluster_default_infraenv }}"
+    import_agents_inventory_path: "/var/secrets/import-agents/servers.yml"
+    import_agents_pull_secret_name: "pull-secret"
+    import_agents_managed_by_label: "osac.openshift.io/managed-by=import-agents"
+
+  tasks:
+    - name: Load server inventory from file
+      ansible.builtin.include_vars:
```
eranco74, line 28: If someone enables `OSAC_IMPORT_AGENTS_ENABLED=true` without creating the `import-agents-inventory` ConfigMap, this `include_vars` will fail with a generic file-not-found error that doesn't hint at the missing ConfigMap.

Consider adding a pre-flight check:
```yaml
    - name: Verify inventory file exists
      ansible.builtin.stat:
        path: "{{ import_agents_inventory_path }}"
      register: _inventory_file
      failed_when: not _inventory_file.stat.exists
```
With a clear error message pointing at the ConfigMap.

tzumainn:
Just a note that we currently have an equivalent role for OpenStack at https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/service/roles/manage_agents/tasks/import_agents.yaml; do we want to make this parallel to that?

tzvatot (commented):
## Re-review

All findings from the previous round were addressed in the merged commit: pre-flight inventory check, configurable per-server BMC TLS verification, explicit `allow_simultaneous: false` / `ask_variables_on_launch: true`, `failed_when` guard on agent labeling, ConfigMap mount at `/var/config/` instead of `/var/secrets/`, and no redundant `when` clauses. Solid turnaround.

| Category | Count |
|----------|-------|
| 🔴 Critical | 0 |
| 🟡 Important | 1 |
| 💡 Suggestion | 1 |

---

### 🟡 Local testing command in docs still won't work

**File:** `docs/import-agents.md:160-162` and `samples/import_agents_extra_vars.yml`

The documented local testing command is:

```bash
ansible-playbook playbook_osac_import_agents.yml \
  -e @samples/import_agents_extra_vars.yml
```

The extra vars file now correctly defines `default_agent_namespace` and `hosted_cluster_default_infraenv` (addressing the original feedback), but it does not override `import_agents_inventory_path`. The playbook will still try to load the server inventory from the default path `/var/config/import-agents/servers.yml`, hit the pre-flight `stat` check, and fail.

The PR body references a different file (`samples/import_agents_servers.yml`) that doesn't exist in the repo, which suggests this was the intended inventory file but got consolidated into the extra vars file.

**Recommendation:** Split `samples/import_agents_extra_vars.yml` into two files — one for the variables (`import_agents_extra_vars.yml`) and one for the inventory (`import_agents_servers.yml`):

```yaml
# samples/import_agents_extra_vars.yml
default_agent_namespace: "hardware-inventory"
hosted_cluster_default_infraenv: "infraenv"
import_agents_inventory_path: "samples/import_agents_servers.yml"
```

```yaml
# samples/import_agents_servers.yml
servers:
  - name: ostest-extraworker-0
    bmc_url: "redfish-virtualmedia+https://..."
    bmc_secret: "ostest-extraworker-0-bmc-secret"
    boot_mac: "00:8f:09:3e:24:af"
    netris_server_name: "extraworker-0"
    resource_class: "x86_64"
  # ...
```

This also makes the sample structure match the production layout (extra vars separate from the inventory ConfigMap content).

---

### 💡 `failed_when` on agent labeling fires after the API call, not before

**File:** `playbook_osac_import_agents.yml:190-194`

The `failed_when` guard correctly catches the case where `mac_to_agent_name` returns `None`, but Ansible evaluates `failed_when` *after* the task executes. So the `kubernetes.core.k8s` module will first attempt to apply a definition with `name: "None"` (the stringified Jinja2 result), get a 404 or validation error from the API server, and *then* `failed_when` marks it failed.

In practice this is harmless — the wait task at line 143 ensures all agents exist before we reach this point, so this is a defense-in-depth guard that should rarely fire. But if it does fire, the user will see two errors (the API error and the `failed_when`), which can be confusing.

A cleaner alternative is a `when` guard that skips + warns instead of failing after-the-fact:

```yaml
      when: >-
        [server.boot_mac] |
        osac.service.mac_to_agent_name(import_agents_final_agents.resources) is not none
      register: import_agents_label_result

    - name: Warn about unlabeled agents
      ansible.builtin.debug:
        msg: "Agent not found for server {{ server.name }} (MAC {{ server.boot_mac }}) — skipping label"
      loop: "{{ import_agents_servers }}"
      loop_control:
        loop_var: server
        label: "{{ server.name }}"
      when: >-
        [server.boot_mac] |
        osac.service.mac_to_agent_name(import_agents_final_agents.resources) is none
```

But the current approach is functional. Up to you.

---

**Overall:** Previous review feedback was fully addressed. The reconciliation logic, stale-resource cleanup ordering, managed-by label scoping, and AAP integration are all clean. The local testing docs are the only actionable item remaining.

PR #284, opened 2026-05-04
Add bare metal fulfillment to CaaS

### LINKED ISSUE #378: Update cluster template(s) to create BareMetalPools for infrastructure setup
Changes may have to be isolated to the "ESI side" of things for now.

ajamias:
## Summary

Use BareMetalPool and HostLease resources to drive agent allocation in cluster_infra create, removing manual NodePool introspection and resource class diffing. Agents are now selected by matching their osac.openshift.io/host_uuid annotation against HostLease externalHostIDs.

Closes osac-project/issues#378

### Important
This PR will break cluster scale-down. One proposed solution is to handle agent scale-down in the profile-specified-template that the host management operator executes for each HostLease. Scale-down will be fixed in a future PR.

## Summary by CodeRabbit

* **Refactor**
  * Updated resource identifiers and label conventions to align with OpenShift infrastructure standards.
  * Simplified bare metal pool creation and host allocation workflows for improved reliability.
  * Streamlined agent selection, attachment, and approval processes with enhanced validation.
  * Replaced dynamic resource inspection with deterministic host lease tracking and matching.

### DIFF
--- a/collections/ansible_collections/massopencloud/esi/plugins/filter/filters.py
+++ b/collections/ansible_collections/massopencloud/esi/plugins/filter/filters.py
@@ -48,14 +48,14 @@ def get_agent_metadata(node_info_list, agents):
                           if port.get('address')]
         agent_name = mac_to_agent_name(node_addresses, agents)
 
-        annotations = {"esi.nerc.mghpcc.org/uuid": node_info['id']}
+        annotations = {"osac.openshift.io/host_uuid": node_info['id']}
 
         topology = extract_esi_location(node_info['name'])
         topology_labels = {"topology.nerc.mghpcc.org/%s" % label: str(val)
                            for label, val in topology.items()
                            if val is not None}
         resource_class_label = {
-            'esi.nerc.mghpcc.org/resource_class':
+            'osac.openshift.io/resource_class':
                 node_info.get('resource_class', '')
         }
         labels = {**topology_labels, **resource_class_label}
--- a/collections/ansible_collections/massopencloud/steps/roles/cluster_infra/tasks/create.yaml
+++ b/collections/ansible_collections/massopencloud/steps/roles/cluster_infra/tasks/create.yaml
@@ -2,6 +2,37 @@
   ansible.builtin.debug:
     var: cluster_infra_node_requests
 
+- name: Build BareMetalPool hostSets
+  ansible.builtin.set_fact:
+    baremetalpool_host_sets: >-
+      {{ baremetalpool_host_sets | default([]) + [{
+        'hostType': item.resourceClass,
+        'replicas': item.numberOfNodes | int
+      }] }}
+  loop: "{{ cluster_infra_node_requests }}"
+
+# For now, we create the BareMetalPool in the POD_NAMESPACE, but
+# in the future, we hope to have it created in the cluster_working_namespace.
+# This requires updates to the bare-metal-fulfillment-operator
+
+- name: Step - Create BareMetalPool
+  kubernetes.core.k8s:
+    state: present
+    definition:
+      apiVersion: osac.openshift.io/v1alpha1
+      kind: BareMetalPool
+      metadata:
+        name: "{{ cluster_infra_name }}"
+        namespace: "{{ lookup('env', 'POD_NAMESPACE') }}"
+        labels: "{{ {cluster_order_label: cluster_infra_name} }}"
+      spec:
+        hostSets: "{{ baremetalpool_host_sets }}"
+  register: baremetalpool_response
+
+- name: Set BareMetalPool UID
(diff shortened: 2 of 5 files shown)


collections/ansible_collections/massopencloud/steps/roles/cluster_infra/tasks/create.yaml
```diff
@@ -73,20 +106,69 @@
...
+    agents_to_attach: []
+
+- name: Select agents matching allocated host IDs
+  ansible.builtin.set_fact:
+    agents_to_attach: "{{ agents_to_attach + [item] }}"
+  loop: "{{ all_agents_result.resources }}"
+  when:
+    - item.metadata.annotations is defined
+    - item.metadata.annotations['esi.nerc.mghpcc.org/uuid'] is defined
+    - item.metadata.annotations['esi.nerc.mghpcc.org/uuid'] in allocated_host_ids
```
tzumainn: I think it might make sense to update this annotation to a more generic name - something like `osac.openshift.io/host_uuid`? The role that imports agents would have to be updated as well.

tzumainn:
Can you also add a note to the PR description that this breaks scale down (but scale up works)?

collections/ansible_collections/massopencloud/steps/roles/cluster_infra/tasks/create.yaml
ajamias: Yep fixed

collections/ansible_collections/massopencloud/steps/roles/cluster_infra/tasks/create.yaml
```diff
@@ -2,6 +2,33 @@
...
+
+- name: Step - Create BareMetalPool
+  kubernetes.core.k8s:
+    state: present
+    definition:
+      apiVersion: osac.openshift.io/v1alpha1
+      kind: BareMetalPool
+      metadata:
+        name: "{{ cluster_infra_name }}"
+        namespace: "{{ lookup('env', 'POD_NAMESPACE') }}"
```
tzumainn, line 26: `cluster_working_namespace` has all the resources associated with the cluster; any reason not to create the BareMetalPool (and the HostLeases) there as well?

collections/ansible_collections/massopencloud/esi/plugins/filter/filters.py
```diff
@@ -48,14 +48,14 @@ def get_agent_metadata(node_info_list, agents):
                           if port.get('address')]
         agent_name = mac_to_agent_name(node_addresses, agents)
 
-        annotations = {"esi.nerc.mghpcc.org/uuid": node_info['id']}
+        annotations = {"osac.openshift.io/host_uuid": node_info['id']}
```
tzumainn, line 51: I think you'll have to update some labels in the `manage_agents` role as well. It probably makes sense to centralize the name of this label in `group_vars/all/osac_common_labels.yaml`

tzumainn (changes requested):
A few more comments; past that it looks good to me. Can you update the PR message to mention that the scale down issue will be fixed in a forthcoming PR?

collections/ansible_collections/massopencloud/steps/roles/cluster_infra/tasks/create.yaml
ajamias: The bare-metal-fulfillment-operator is installed in the namespace with all the other operators (fulfillment-operator, osac-operator). Should it still go into the `cluster_working_namespace`?

tzumainn: Huh, I guess the OSAC operators must be configured to only reconcile objects in their installed namespace? On the other hand, the hosted cluster operator covers all namespaces. Is there a clever way you can think of to have the OSAC operators reconcile these cluster working namespaces as well? That would be ideal, since theoretically the cluster working namespace contains all resources used for creation of the cluster.

ajamias: So for now, we will have it be made in the `POD_NAMESPACE`, but later we will have to update the `bare-metal-fulfillment-operator` to support watching over its resources from different namespaces.

tzumainn: Yep; can you leave a comment in the code to that effect?

tzumainn (changes requested):
Just noticed that the PR message still refers to the old host UUID label; can you update it to the new one (and explicitly mention that scale down will be fixed in a future PR)?

tzumainn (approved):
Thanks - looks good to me!

PR #287, opened 2026-05-05
Initial OpenStack network implementation

tzumainn:
Adds an OpenStack network implementation that supports VirtualNetworks and Subnets. Many OpenStack-specific settings are currently defaulted; once `template_parameters` are supported, these settings can be read from there.

### DIFF
--- a/collections/ansible_collections/osac/templates/roles/openstack/README.md
+++ b/collections/ansible_collections/osac/templates/roles/openstack/README.md
@@ -1,0 +1,116 @@
+# openstack
+
+Provisions networking resources using OpenStack Neutron.
+
+## Resources
+
+### VirtualNetwork
+
+VirtualNetworks define the top-level network isolation boundary with CIDR allocation and implementation strategy selection via NetworkClass.
+
+**Key behaviors:**
+- Creates Neutron network resource in OpenStack
+- Network MTU defaults to 1500 (configurable via `default_network_mtu`)
+- Networks are created as internal (non-external) networks
+- One VirtualNetwork maps to one Neutron network
+- Network name matches VirtualNetwork CR name
+
+**Implementation:**
+- Uses `openstack.cloud.network` module
+- Network properties configured from defaults (MTU)
+- Network deletion is idempotent (no error if already deleted)
+
+### Subnet
+
+Subnets subdivide VirtualNetworks into logical segments with DHCP, DNS, and IP allocation.
+
+**Key behaviors:**
+- Creates Neutron subnet within the parent VirtualNetwork's network
+- IP allocation pool reserves first 10 and last 10 IPs in CIDR range
+- Gateway set to first usable IP in CIDR
+- DNS nameservers default to `8.8.8.8` (configurable via `default_subnet_dns_nameservers`)
+- Optional router creation for external connectivity (controlled by `default_subnet_external_gateway`)
+- One Subnet maps to one Neutron subnet (plus optional router)
+
+**IP Allocation:**
+- For `10.0.0.0/24`:
+  - Gateway: `10.0.0.1`
+  - Allocation pool: `10.0.0.10` - `10.0.0.245`
+  - Reserved: first 9 IPs and last 10 IPs
+
+**Router behavior:**
+- When `default_subnet_external_gateway` is `true`, creates router named `router-{subnet-name}`
+- Router connects subnet to external network (default: `external`, configurable via `default_subnet_external_network`)
+- Router deleted automatically when subnet is deleted
+
+**Fallback CIDR:**
+- If VirtualNetwork or Subnet does not specify `ipv4Cidr`, uses `default_subnet_ipv4_cidr` (default: `192.168.50.0/24`)
+
+### SecurityGroup
+
+**Not yet implemented.**
+
+## Implementation Strategy
+
+This role implements the `openstack` NetworkClass strategy using OpenStack Neutron. The implementation follows these patterns:
+
+**For VirtualNetworks:**
+- Create Neutron network with MTU configuration
+- Network name matches CR name for idempotent operations
+
+**For Subnets:**
+- Create Neutron subnet within parent network
+- Configure DHCP, DNS, and allocation pools
+- Optionally create router for external connectivity
+
+**For SecurityGroups:**
+- Planned: will translate to Neutron security groups and rules
+
+## Task Files
+
+- `tasks/create_virtual_network.yaml` - Creates Neutron network from VirtualNetwork resource
+- `tasks/delete_virtual_network.yaml` - Removes Neutron network
+- `tasks/create_subnet.yaml` - Creates Neutron subnet and optional router from Subnet resource
+- `tasks/delete_subnet.yaml` - Removes router and Neutron subnet
+
+## Usage
+
+### Example: VirtualNetwork Provisioning
+
+```yaml
+- name: Create VirtualNetwork
+  ansible.builtin.include_role:
+    name: openstack
+    tasks_from: create_virtual_network
+  vars:
+    virtual_network: "{{ ansible_eda.event.payload }}"
+    virtual_network_name: "{{ ansible_eda.event.payload.metadata.name }}"
+```
+
+### Example: Subnet Provisioning
+
+```yaml
+- name: Create Subnet
+  ansible.builtin.include_role:
+    name: openstack
+    tasks_from: create_subnet
+  vars:
+    subnet: "{{ ansible_eda.event.payload }}"
+    subnet_name: "{{ ansible_eda.event.payload.metadata.name }}"
+```
+
+### Example: Custom Subnet Configuration
+
+```yaml
+- name: Create Subnet with custom settings
+  ansible.builtin.include_role:
+    name: openstack
+    tasks_from: create_subnet
+  vars:
+    subnet: "{{ ansible_eda.event.payload }}"
+    subnet_name: "{{ ansible_eda.event.payload.metadata.name }}"
+    default_subnet_dns_nameservers:
+      - 10.0.0.10
+      - 10.0.0.11
+    default_subnet_external_gateway: false
+```
--- a/collections/ansible_collections/osac/templates/roles/openstack/meta/osac.yaml
+++ b/collections/ansible_collections/osac/templates/roles/openstack/meta/osac.yaml
@@ -1,0 +1,13 @@
+---
+title: OpenStack Neutron Network Implementation
+description: >
+  Provisions networking resources using OpenStack Neutron
+
+template_type: network
+
+# NetworkClass registration fields
+implementation_strategy: openstack
+capabilities:
+  supports_ipv4: true
+  supports_ipv6: false
+  supports_dual_stack: false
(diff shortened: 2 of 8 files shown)


tzumainn:
@adriengentil does this implementation look right to you?

tzvatot (commented):
<removing comment in favor of inline comments> (BTW why can't I delete my own comment?)

collections/ansible_collections/osac/templates/roles/openstack/tasks/delete_subnet.yaml
```diff
@@ -0,0 +1,19 @@
...
+    sn_name: "{{ subnet.metadata.name }}"
+
+- name: Delete router
+  openstack.cloud.router:
+    name: "router-{{ sn_name }}"
+    state: absent
+
+- name: Delete subnet
+  openstack.cloud.subnet:
+    name: "{{ sn_name }}"
```
tzvatot, line 13: 🔴 **Delete subnet doesn't clean up the router's port before deleting the subnet**

The delete task deletes the router first, then the subnet. But `openstack.cloud.router` with `state: absent` deletes the router entirely — including for other subnets that may share it. More critically, if the router has an interface on the subnet, Neutron will refuse to delete the subnet until the port is removed. The create path adds the subnet as a router interface, so the delete path should remove the interface first:

```yaml
- name: Remove subnet from router
  openstack.cloud.router:
    name: "router-{{ sn_name }}"
    interfaces: []
  # Remove interface before deleting

- name: Delete router
  openstack.cloud.router:
    name: "router-{{ sn_name }}"
    state: absent
```

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_subnet.yaml
```diff
@@ -0,0 +1,40 @@
...
+  when: default_subnet_external_gateway
+  openstack.cloud.router:
+    name: "router-{{ sn_name }}"
+    network: "{{ default_subnet_external_network }}"
+    interfaces:
+      - "{{ subnet_result.id }}"
+  register: router_result
+
+- name: Display OpenStack router creation result
+  ansible.builtin.debug:
```
tzvatot, line 55: 🟡 **Router creation result logged with wrong variable**

This logs `subnet_result.id` instead of `router_result.id`. Should be:

```yaml
msg:
  - "OpenStack router for subnet '{{ sn_name }}' created successfully"
  - "Resource UID: {{ router_result.id }}"
```

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_subnet.yaml
```diff
@@ -0,0 +1,40 @@
...
+- name: Extract Subnet configuration
+  ansible.builtin.set_fact:
+    sn_name: "{{ subnet.metadata.name }}"
+    sn_ipv4_cidr: "{{ subnet.spec.ipv4Cidr | default(default_subnet_ipv4_cidr) }}"
+    sn_virtual_network: "{{ subnet.spec.virtualNetwork }}"
+
+- name: Create subnet
+  openstack.cloud.subnet:
+    name: "{{ sn_name }}"
+    network: "{{ sn_virtual_network }}"
```
tzvatot, line 31: 🟡 **Hardcoded allocation pool offsets**

`nthhost(10)` and `nthhost(-10)` hardcode the allocation pool boundaries. For small CIDRs (e.g., /28 = 16 addresses), this leaves zero usable IPs. These should either be configurable via defaults or validated against the CIDR size.

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_subnet.yaml
```diff
@@ -0,0 +1,40 @@
...
+- name: Display OpenStack subnet creation result
+  ansible.builtin.debug:
+    msg:
+      - "OpenStack subnet '{{ sn_name }}' created successfully"
+      - "Resource UID: {{ subnet_result.id }}"
+
+- name: Create router
+  when: default_subnet_external_gateway
+  openstack.cloud.router:
+    name: "router-{{ sn_name }}"
```
tzvatot, line 48: 💡 **No idempotency handling on router interface attachment**

The router creation may fail if re-run because the interface is already attached. Consider adding `ignore_errors` or a pre-check for the router interface step.

collections/ansible_collections/osac/templates/roles/openstack/README.md
```diff
@@ -0,0 +1,65 @@
...
+### VirtualNetwork
+
+
+### Subnet
+
+
+### SecurityGroup
+
+
+## Implementation Strategy
```
tzvatot, line 53: 🟡 **README sections are empty**

VirtualNetwork, Subnet, SecurityGroup, and Implementation Strategy sections have headers but no content. Is that WIP?

tzvatot (changes requested):

collections/ansible_collections/osac/templates/roles/openstack/tasks/delete_subnet.yaml
tzumainn: The router does have an interface on the subnet; however, from my testing deleting the router will also remove taht interface.

There should be no way for other subnets to use this router; it's created through the template mechanism as well, and it is not exposed through OSAC in any way.

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_subnet.yaml
tzumainn: whoops - fixed!

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_subnet.yaml
larsks: I think both: `allocation_pool_start` and `allocation_pool_end` should be configurable, but we should also validate them against the CIDR size so that if someone does select a small cidr range without changing the defaults we provide a meaningful error.

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_subnet.yaml
tzumainn: That's not what happens; if you re-run the creation multiple times and the router has already been created, then openstack simply won't re-run the creation.

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_subnet.yaml
larsks: Here's a thought on validation; we calculate the number of addresses in the allocation pool and fail if it is <= 0.

```yaml
- name: Calculate pool_start and pool_end
  set_fact:
    pool_start: "{{sn_ipv4_cidr | ansible.utils.nthhost(10)}}"
    pool_end: "{{sn_ipv4_cidr | ansible.utils.nthhost(-10)}}"

- name: Calculate pool size
  set_fact:
    pool_size: >
      {{
        (pool_end | ansible.utils.ipaddr('int')) - (pool_start |
        ansible.utils.ipaddr('int')) + 1
      }}

- name: Validate pool size
  assert:
    that: pool_size > 0
    msg: "Allocation pool size {{ pool_size }} is invalid for cidr {{ sn_ipv4_cidr }}"
```

vendor/ansible_collections/nvidia/bare_metal/plugins/modules/allocation_constraint.py
```diff
@@ -30,30 +30,32 @@
   constraint_type:
     type: str
     description:
-    - constraint_type parameter.
+    - Type of the Allocation Constraint. Please note that OnDemand and Preemptible are not supported by current implementation.
     choices:
     - Reserved
     - OnDemand
     - Preemptible
```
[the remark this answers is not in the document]
danmanor: I guess the `vendor/` changes are not intentional

danmanor (changes requested):

collections/ansible_collections/osac/templates/roles/openstack/README.md
tzumainn: Good call - I've updated this README!

collections/ansible_collections/osac/templates/roles/openstack/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+
+## Task Files
+
+- `tasks/create_virtual_network.yaml` - Creates ClusterUserDefinedNetwork CR from VirtualNetwork resource
+- `tasks/delete_virtual_network.yaml` - Removes ClusterUserDefinedNetwork CR
+- `tasks/create_subnet.yaml` - Creates namespace with CUDN labels from Subnet resource
+- `tasks/delete_subnet.yaml` - Removes namespace
+- `tasks/create_security_group.yaml` - Creates NetworkPolicy resources from SecurityGroup rules
+- `tasks/delete_security_group.yaml` - Removes NetworkPolicy resources
```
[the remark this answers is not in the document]
tzumainn: Removed!

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_virtual_network.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: Extract VirtualNetwork configuration
+  ansible.builtin.set_fact:
+    vn_name: "{{ virtual_network.metadata.name }}"
+
+- name: Create network
+  openstack.cloud.network:
+    name: "{{ vn_name }}"
+    external: "false"
+    mtu: "{{ default_network_mtu }}"
```
larsks, line 12: The MTU should be configurable per-network.

tzumainn:
@danmanor yep, including the vendor files was an accident - they should be removed now!

collections/ansible_collections/osac/templates/roles/openstack/meta/osac.yaml
```diff
@@ -0,0 +1,13 @@
...
+  Provisions networking resources using OpenStack Neutron
+
+template_type: network
+
+# NetworkClass registration fields
+implementation_strategy: openstack
+capabilities:
+  supports_ipv4: true
+  supports_ipv6: true
+  supports_dual_stack: true
```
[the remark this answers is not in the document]
tzumainn: Good call - updated to set `supports_ipv6` and `supports_dual_stack` to false.

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_virtual_network.yaml
tzumainn: Agreed, but MTU isn't currently exposed in the VirtualNetwork CR; I believe the idea was to default this for now until it was.

collections/ansible_collections/osac/templates/roles/openstack/tasks/create_subnet.yaml
tzumainn: Good idea (and thanks for the code snippet)! Updated, and also changed the pool start/end offsets into defaults.

tzumainn:
@tzvatot @larsks thanks for the reviews! I believe I've addressed all the comments; let me know if I missed something!

tzvatot (approved):
/lgtm

PR #291, opened 2026-05-07
OSAC-175, OSAC-176: Tier-aware StorageClass role and template-driven selection

zszabo-rh:
## Summary

- Update `tenant_storage_class` role to read the resolved `storageClasses` list from extra_vars (injected by the CI controller) instead of querying the Tenant CR via `k8s_info`
- Require `tenant_storage_class_storage_tier` parameter with no default value — templates must be explicit about which tier they need
- Move `tenant_storage_class` role invocation from playbook `pre_tasks` into the `ocp_virt_vm` template role, which passes `tier: "default"`
- Playbook extracts `tenant_storage_classes` from `ansible_eda.event.tenant_storage_classes` and sets it as a fact for downstream roles

Per the [Tenant Storage Tiers](https://github.com/osac-project/enhancement-proposals/blob/main/enhancements/tenant-storage-tiers/README.md) enhancement proposal, templates use template-driven tier selection: each template hardcodes which tier its disks need, and the user doesn't need to know about storage tiers.

### Flow

```
CI controller injects tenant_storage_classes into extra_vars (OSAC-687)
  → Playbook extracts it as a variable
  → Template role (ocp_virt_vm) calls tenant_storage_class with tier: "default"
  → Role matches tier in the list → sets tenant_storage_class_name
  → Template uses it for DataVolume creation
```

## Test plan

- [x] `ansible-lint` passes
- [x] Role tests are self-contained (no cluster needed): happy path, tier not available, empty list, duplicate tier first-match

Jira: [OSAC-175](https://redhat.atlassian.net/browse/OSAC-175), [OSAC-176](https://redhat.atlassian.net/browse/OSAC-176)

**Companion PR:** [osac-operator#229](https://github.com/osac-project/osac-operator/pull/229) (OSAC-687) injects `tenant_storage_classes` into extra_vars from the CI controller.

### DIFF
--- a/playbook_osac_create_compute_instance.yml
+++ b/playbook_osac_create_compute_instance.yml
@@ -27,9 +27,7 @@
           ComputeInstance '{{ compute_instance_name }}' has no tenant_storage_classes
           in extra_vars. The osac-operator CI controller should inject the resolved
           storageClasses list before triggering provisioning.
-      when: >-
-        tenant_storage_classes is not defined or
-        tenant_storage_classes | length == 0
+      when: tenant_storage_classes | length == 0
 
     - name: Set storage classes list for downstream roles
       ansible.builtin.set_fact:
--- a/playbook_osac_create_compute_instance.yml
+++ b/playbook_osac_create_compute_instance.yml
@@ -29,9 +29,10 @@
           storageClasses list before triggering provisioning.
       when: tenant_storage_classes | length == 0
 
-    - name: Set storage classes list for downstream roles
+    - name: Set storage classes context for downstream roles
       ansible.builtin.set_fact:
         tenant_storage_class_storage_classes: "{{ tenant_storage_classes }}"
+        tenant_storage_class_tenant_name: "{{ compute_instance.status.tenantReference.name }}"
 
   tasks:
 
--- a/playbook_osac_create_compute_instance.yml
+++ b/playbook_osac_create_compute_instance.yml
@@ -8,6 +8,7 @@
     compute_instance_name: "{{ ansible_eda.event.payload.metadata.name }}"
     template_id: "{{ ansible_eda.event.payload.spec.templateID }}"
     template_parameters: {}
+    tenant_storage_classes: "{{ ansible_eda.event.tenant_storage_classes | default([]) }}"
 
   pre_tasks:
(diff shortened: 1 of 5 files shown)


akshaynadkarni (commented):
(Moved to inline comment on playbook_osac_create_compute_instance.yml line 11)

playbook_osac_create_compute_instance.yml
```diff
@@ -8,6 +8,7 @@
     compute_instance_name: "{{ ansible_eda.event.payload.metadata.name }}"
     template_id: "{{ ansible_eda.event.payload.spec.templateID }}"
     template_parameters: {}
+    tenant_storage_classes: "{{ ansible_eda.event.tenant_storage_classes | default([]) }}"
```
akshaynadkarni, line 11: I raised a question with the team about whether we still need to support EDA as a provisioning provider: [slack thread](https://redhat-internal.slack.com/archives/C08ESMFV85Q/p1778614449356309) and cc'd you for reference.

The `tenant_storage_classes` injection via extra_vars only works with the AAP direct provider. If someone configures `OSAC_PROVISIONING_PROVIDER=eda`, this key won't exist in the event payload, the `default([])` fallback will produce an empty list, and line 30 will fail the playbook.

If the team decides we need to continue supporting EDA, we should either:
- Keep the old `k8s_info` lookup as a fallback when `tenant_storage_classes` is empty, or
- Hardcode `NewAAPProvider` for ComputeInstance in the operator (the same way networking already does), so CI always uses AAP direct regardless of the env var

zszabo-rh: I know, this was intentional.. it was designed for AAP direct, which I think is the direction the project is heading, since networking already hardcodes it regardless of the env var.            

Based on the Slack thread, it looks like EDA removal is more or less already agreed. Once confirmed, this concern goes away and the follow-up to flatten the `ansible_eda.event.payload` wrapper into top-level `extra_vars` keys is a separate cleanup since it affects all playbooks.

akshaynadkarni: Totally agree. I'm glad to see the team is agreeing to removing support for EDA.

akshaynadkarni (approved):
LGTM

PR #301, opened 2026-05-18
Add HostLease playbooks, bm_host_agent_provisioning role, and AAP config

DanNiESh:
## Summary

   - Add `bm_host_agent_provisioning` and `bm_host_agent_deprovisioning` template roles that provision bare metal nodes via OpenStack Ironic ramdisk deploy (create), detach neutron ports and undeploy them on deletion (delete)
   - Add `playbook_osac_create_host_lease.yml` and `playbook_osac_delete_host_lease.yml` that extract HostLease CR fields and delegate to the template role based on `spec.templateID`
   - Register AAP config-as-code resources: bare-metal-fulfillment inventory, inventory source, instance group (with
    OpenStack `clouds.yaml` mount), and job templates for create/delete

   ## Details

   - The provisioning role uses `openstack.cloud.baremetal_node_action` to deploy nodes with a `boot_iso` via the
   ramdisk deploy interface
   - The deprovisioning role detaches all neutron ports via `massopencloud.esi.l2` before undeploying the node
   - The instance group pod spec mounts `osac-os-clouds` secret at `/etc/openstack/clouds.yaml` and sets
   `OS_CLIENT_CONFIG_FILE` so the OpenStack SDK can authenticate to Ironic
   - Environment variables from `bare-metal-fulfillment-ig` secret are injected via `envFrom`

### DIFF
--- a/collections/ansible_collections/osac/config_as_code/roles/aap/vars/controller.yml
+++ b/collections/ansible_collections/osac/config_as_code/roles/aap/vars/controller.yml
@@ -279,6 +279,30 @@ controller_templates: # noqa: var-naming[no-role-prefix]
     allow_simultaneous: false
     ask_variables_on_launch: true
     verbosity: 0
+  - name: "{{ aap_prefix }}-create-host-lease"
+    project: "{{ aap_prefix }}"
+    organization: "{{ aap_organization_name }}"
+    job_type: run
+    playbook: "playbook_osac_create_host_lease.yml"
+    inventory: "{{ aap_prefix }}-bare-metal-fulfillment"
+    execution_environment: "{{ aap_prefix }}-ee"
+    instance_groups:
+      - "{{ aap_prefix }}-bare-metal-fulfillment-ig"
+    allow_simultaneous: true
+    ask_variables_on_launch: true
+    verbosity: 0
+  - name: "{{ aap_prefix }}-delete-host-lease"
+    project: "{{ aap_prefix }}"
+    organization: "{{ aap_organization_name }}"
+    job_type: run
+    playbook: "playbook_osac_delete_host_lease.yml"
+    inventory: "{{ aap_prefix }}-bare-metal-fulfillment"
+    execution_environment: "{{ aap_prefix }}-ee"
+    instance_groups:
+      - "{{ aap_prefix }}-bare-metal-fulfillment-ig"
+    allow_simultaneous: true
+    ask_variables_on_launch: true
+    verbosity: 0
 
 controller_job_template_surveys: # noqa: var-naming[no-role-prefix]
   - name: "{{ aap_prefix }}-create-hosted-cluster-post-install"
@@ -380,6 +404,9 @@ controller_inventories: # noqa: var-naming[no-role-prefix]
   - name: "{{ aap_prefix }}-networking-operations"
     description: "Networking Operational Inventory"
     organization: "{{ aap_organization_name }}"
+  - name: "{{ aap_prefix }}-bare-metal-fulfillment"
+    description: "Bare Metal Fulfillment Inventory"
+    organization: "{{ aap_organization_name }}"
 
 controller_inventory_sources: # noqa: var-naming[no-role-prefix]
   - name: "{{ aap_prefix }}-cluster-fulfillment-is"
@@ -424,6 +451,15 @@ controller_inventory_sources: # noqa: var-naming[no-role-prefix]
     source_project: "{{ aap_prefix }}"
     source_path: "inventory/localhost.yml"
     inventory: "{{ aap_prefix }}-networking-operations"
+    overwrite: true
+    overwrite_vars: true
+    update_cache_timeout: 0
+  - name: "{{ aap_prefix }}-bare-metal-fulfillment-is"
+    organization: "{{ aap_organization_name }}"
+    source: scm
+    source_project: "{{ aap_prefix }}"
+    source_path: "inventory/localhost.yml"
+    inventory: "{{ aap_prefix }}-bare-metal-fulfillment"
     overwrite: true
     overwrite_vars: true
     update_cache_timeout: 0
@@ -688,3 +724,81 @@ controller_instance_groups: # noqa: var-naming[no-role-prefix]
   - name: "{{ aap_prefix }}-networking-operations-ig"
     is_container_group: true
     pod_spec_override: "{{ lookup('ansible.builtin.template', 'networking-operations-ig.j2') }}"
+
+  - name: "{{ aap_prefix }}-bare-metal-fulfillment-ig"
+    is_container_group: true
+    pod_spec_override: |
+      apiVersion: v1
+      kind: Pod
+      metadata:
+        labels:
+          ansible_job: ''
+      spec:
+        serviceAccountName: osac-sa
+        affinity:
+          podAntiAffinity:
+            preferredDuringSchedulingIgnoredDuringExecution:
+              - weight: 100
+                podAffinityTerm:
+                  labelSelector:
+                    matchExpressions:
+                      - key: ansible_job
+                        operator: Exists
+                  topologyKey: kubernetes.io/hostname
+        containers:
+          - image: >-
+              registry.redhat.io/ansible-automation-platform-25/ee-supported-rhel8@sha256:d8400a472e769d0f3d591dafaad318522009c583b08e881c23b6d57a27cc10ed
+            name: worker
+            args:
+              - ansible-runner
+              - worker
+              - '--private-data-dir=/runner'
+            volumeMounts:
+              - name: kube-api-access
+                mountPath: /var/run/secrets/kubernetes.io/serviceaccount
+                readOnly: true
+              - name: os-clouds
+                mountPath: /etc/openstack
+                readOnly: true
+            env:
+              - name: POD_NAMESPACE
+                valueFrom:
+                  fieldRef:
+                    fieldPath: metadata.namespace
+              - name: OS_CLIENT_CONFIG_FILE
+                value: /etc/openstack/clouds.yaml
+            envFrom:
+              - secretRef:
+                  name: bare-metal-fulfillment-ig
+                  optional: true
+        volumes:
+          - name: kube-api-access
+            projected:
+              sources:
+                - serviceAccountToken:
+                    path: token
+                    expirationSeconds: 3600
+                - configMap:
+                    name: kube-root-ca.crt
+                    items:
+                      - key: ca.crt
+                        path: ca.crt
+                - downwardAPI:
+                    items:
+                      - path: namespace
+                        fieldRef:
+                          apiVersion: v1
+                          fieldPath: metadata.namespace
+                - configMap:
+                    name: openshift-service-ca.crt
+                    items:
+                      - key: service-ca.crt
+                        path: service-ca.crt
+              defaultMode: 420
+          - name: os-clouds
+            secret:
+              secretName: osac-os-clouds
+              items:
+                - key: clouds.yaml
+                  path: clouds.yaml
+              defaultMode: 420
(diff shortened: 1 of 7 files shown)


tzumainn:
I'd suggest either including `openstack` in the name `bm_host_image_provision`, or updating the role to fail if the host lease doesn't have a host class of `openstack`

collections/ansible_collections/osac/templates/roles/bm_host_image_provision/tasks/create.yaml
```diff
@@ -0,0 +1,51 @@
+---
+- name: Skip provisioning when state is not active
+  when: provisioning_state != "active"
+  ansible.builtin.debug:
+    msg: "Provisioning state is '{{ provisioning_state }}', skipping image provision"
```
tzumainn: I'd suggest using `host_lease.spec.provisioning` instead of `provisioning_state`; I know we're not explicitly "passing in" any parameters into this role, but pretending that it's simply the host lease makes it far clearer where values like the provisioning state are actually coming from.

You can see an example of this at https://github.com/osac-project/osac-aap/blob/main/collections/ansible_collections/osac/templates/roles/ocp_4_17_small/tasks/install.yaml and the use of `cluster_order`

playbook_osac_create_host_lease.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Create a HostLease resource
```
tzumainn: This description seems slightly off; we're not creating a host lease resource with these templates, we're configuring existing host lease resources

playbook_osac_create_host_lease.yml
```diff
@@ -0,0 +1,30 @@
...
+    provisioning_network: "{{ provisioning.provisioningNetwork | default('') }}"
+
+  pre_tasks:
+    - name: Display host lease information
+      ansible.builtin.debug:
+        msg:
+          - "Host: {{ host_lease_name }} ({{ host_lease_namespace }})"
+          - "External Host ID: {{ external_host_id }}"
+          - "Provisioning State: {{ provisioning_state }}"
+          - "Image URL: {{ image_url }}"
```
tzumainn: I don't know that we have to have the provisioning state and image URL here; in a general sense, they're irrelevant when running a generic host lease template

tzumainn (changes requested):

playbook_osac_create_host_lease.yml
```diff
@@ -0,0 +1,30 @@
...
+
+  tasks:
+    - name: Call the selected host lease template
+      ansible.builtin.include_role:
+        name: "osac.templates.{{ template_id }}"
+        tasks_from: create
```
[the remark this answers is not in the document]
DanNiESh: Prefer not using an allowlist for templateID to be consistent with osac-aap pattern.

ajamias:
> I'd suggest either including `openstack` in the name `bm_host_image_provision`, or updating the role to fail if the host lease doesn't have a host class of `openstack`

This assumes that we will have a role for each inventory management implementation (openstack, etc). Do we want that or do we want to specify a general action such as image provisioning and specify the implementation in the task files?

tzumainn:
> > I'd suggest either including `openstack` in the name `bm_host_image_provision`, or updating the role to fail if the host lease doesn't have a host class of `openstack`
> 
> This assumes that we will have a role for each inventory management implementation (openstack, etc). Do we want that or do we want to specify a general action such as image provisioning and specify the implementation in the task files?

Good question, and I guess we should avoid a role per implementation. So I guess I'd lean towards the second option (updating the role to fail if the host lease doesn't have a host class of `openstack`).

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
```diff
@@ -0,0 +1,51 @@
...
+                | ternary({'deploy_interface': 'ramdisk'}, {})
+              )
+          }}
+
+    - name: Deploy bare metal node with ramdisk image
+      openstack.cloud.baremetal_node_action:
+        name: "{{ external_host_id }}"
+        instance_info: "{{ node_instance_info }}"
+        deploy: true
+        wait: "{{ bm_deploy_wait }}"
```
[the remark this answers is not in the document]
DanNiESh: Done.

playbook_osac_create_host_lease.yml
DanNiESh: Fixed

collections/ansible_collections/osac/templates/roles/bm_host_image_provision/tasks/create.yaml
DanNiESh: Updated the create/delete tasks to read parameters from host_lease directly

playbook_osac_create_host_lease.yml
DanNiESh: Removed all irrelevant parameters in the general host lease template

collections/ansible_collections/osac/config_as_code/roles/aap/vars/controller.yml
```diff
@@ -688,3 +724,81 @@ controller_instance_groups: # noqa: var-naming[no-role-prefix]
...
+                    items:
+                      - key: service-ca.crt
+                        path: service-ca.crt
+              defaultMode: 420
+          - name: os-clouds
+            secret:
+              secretName: osac-os-clouds
+              items:
+                - key: clouds.yaml
+                  path: clouds.yaml
```
ajamias, line 803: The cluster-fulfillment-ig imports OpenStack credentials via env variables. Should we do the same for bare-metal-fulfillment-ig?

tzumainn: Good point - yeah, consistency is probably ideal.

tzumainn: haha j/k

tzumainn (commented):
Thinking about it: we talked about how we might want to separate out the `bm_host_image_provision` role because it's being used for two separate profiles - agent provisioning, and agent deprovisioning. To make that clear, I think I'd call the template `bm_host_agent_provisioning` (and create the new `bm_host_agent_deprovisioning`), and perhaps put a debug message in the "blank" templates of each. For example, for `bm_host_agent_provisioning`, the `delete.yaml` playbook might simply have a debug statement that states that the host is now ready to be used as an agent.

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - name: Deploy bare metal node with ramdisk image
+      openstack.cloud.baremetal_node_action:
+        name: "{{ host_lease.spec.externalHostID }}"
+        instance_info: "{{ node_instance_info }}"
+        deploy: true
+        wait: "{{ bm_deploy_wait }}"
+
+    - name: Image provisioning completed
+      ansible.builtin.debug:
+        msg: "Image provisioning completed for node {{ host_lease.spec.externalHostID }}"
```
tzumainn: One step missing here: after the node is provisioned, we want to put it on a specific network (in our case, it's the `idle-agents-network`). The network is presumably a template parameter.

I know we've been trying to avoid using the direct openstack commands, but I think using `massopencloud/esi/roles/l2/tasks/set_networks_for_node.yaml` for now should be fine

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
```diff
@@ -0,0 +1,65 @@
...
+- name: Fail if host class is not openstack
+  ansible.builtin.fail:
+    msg: "bm_host_agent_provisioning requires hostClass 'openstack', got '{{ host_lease.spec.hostClass | default('') }}'"
+  when: host_lease.spec.hostClass | default('') != "openstack"
+
+- name: Parse template parameters
+  ansible.builtin.set_fact:
+    template_params: "{{ host_lease.spec.templateParameters | from_json }}"
+
+- name: Skip provisioning when state is not active
```
tzumainn: I don't think we need this; if this template is being called, then presumably the user wants this host provisioned and should not have to supply a provisioningState parameter.

We *should* have a check to see whether the node is already provisioned.

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
DanNiESh: Image provisioning also need the node to have a neutron port. Should we put the node on `idle-agents-network` before or after node is provisioned?

tzumainn: Doesn't the node have to be on the provisioning network during provisioning?

DanNiESh: Yes. So we will provide 2 network parameters in `templateParams`: one is `provisioningNetwork` the other one is `hostNetwork` (or something similar)?

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
DanNiESh: Added a check to see whether the node is `active`. Also added a task to move node to `available` state before provisioning if it's in other states. Question: if the node is already provisioned, do we want to undeploy and re-provision the node since it might not be provisioned with the desired image?

tzumainn: That's a good point. These tasks should be idempotent, so I think if the node is already provisioned, you'd want to:

* check if it already specifies the desired image; if so, then you can assume it's already provisioned correctly
* otherwise, undeploy and re-provision as you say

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
tzumainn: I'd say the provisioning network could be a default; I think it's a uniquely Ironic thing. I'd call the other one `agentPrivateNetwork` just to make it explicit (and since the template should be agent-specific).

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
```diff
@@ -0,0 +1,78 @@
+---
+- name: Fail if host class is not openstack
+  ansible.builtin.fail:
+    msg: "bm_host_agent_provisioning requires hostClass 'openstack', got '{{ host_lease.spec.hostClass | default('') }}'"
+  when: host_lease.spec.hostClass | default('') != "openstack"
+
+- name: Parse template parameters
+  ansible.builtin.set_fact:
+    template_params: "{{ host_lease.spec.templateParameters | from_json }}"
```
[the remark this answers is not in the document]
DanNiESh: if the node is active, we still need `templateParameters` to check if the image url and network match or not

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
DanNiESh: Done! Calling `massopencloud/esi/roles/l2/tasks/set_networks_for_node.yaml` to attach node to provisioning network before node deploy. And attach the node to `agentPrivateNetwork` after it's provisioned.

collections/ansible_collections/osac/templates/roles/bm_host_agent_provisioning/tasks/create.yaml
DanNiESh: Done!

DanNiESh:
Everything's been addressed and it is ready for re-review.

tzumainn (approved):
/lgtm

PR #304, opened 2026-05-18
fix: HyperConverged GPU passthrough path for `hco.kubevirt.io/v1`

tchughesiv:
## Summary

- Fixes GPU passthrough registration on clusters using HyperConverged `hco.kubevirt.io/v1`, where `permittedHostDevices` lives under `spec.virtualization` instead of `spec.permittedHostDevices`.
- Updates `configure_permitted_host_devices.yaml` to read/patch the v1 path with a v1beta1 fallback for older CNV installs.
- Updates the `compute_instance_with_gpu_create` integration test to read the v1 CR and assert against the resolved `pciHostDevices` list.

_failure example - https://github.com/osac-project/osac-aap/actions/runs/26055118992/job/76601196240_

## Background

PR #258 added automatic HyperConverged registration of permitted PCI host devices for GPU passthrough. The role and integration test assumed the `v1beta1` schema. Current HCO CRDs (including the kind test cluster) store `HyperConverged` as `v1`, so patches targeted the wrong field and `compute_instance_with_gpu_create:baseline` failed even though VM `hostDevices` were configured correctly.

### DIFF
--- a/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/meta/argument_specs.yaml
+++ b/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/meta/argument_specs.yaml
@@ -17,7 +17,8 @@ argument_specs:
           in the KubeVirt domain devices spec, and will also ensure that the
           `HyperConverged` CR (`kubevirt-hyperconverged` in the `openshift-cnv`
           namespace) permits these host devices under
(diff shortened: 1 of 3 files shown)


jhernand (approved):

akshaynadkarni (approved):
LGTM

PR #310, opened 2026-05-20
Add BareMetalPool Playbooks, bm_private_network role, and AAP config

ajamias:
Adds playbooks to run on BareMetalPool create and delete for Profiles. Creates an initial template that creates a private network.

Depends on #301

### DIFF
--- a/collections/ansible_collections/osac/templates/roles/bm_private_network/defaults/main.yaml
+++ b/collections/ansible_collections/osac/templates/roles/bm_private_network/defaults/main.yaml
@@ -1,0 +1,6 @@
+---
+# We define default_network_class as a temporary workaround the problem that
+# BareMetalPools have no idea what network class to use. The solution will be
+# to specify the network class in the BareMetalPool's status and that will be
+# a future change in the bare-metal-fulfillment-operator.
+default_network_class: openstack
--- a/collections/ansible_collections/osac/templates/roles/bm_private_network/tasks/create.yaml
+++ b/collections/ansible_collections/osac/templates/roles/bm_private_network/tasks/create.yaml
@@ -1,0 +1,7 @@
+---
+- name: Extract network class from BareMetalPool status
+  ansible.builtin.set_fact:
+    network_class: "{{ bare_metal_pool.status.networkClass | default(default_network_class) }}"
+
+- name: Create private network using network class implementation
+  ansible.builtin.include_tasks: "create_{{ network_class }}.yaml"
--- a/playbook_osac_create_bare_metal_pool.yml
+++ b/playbook_osac_create_bare_metal_pool.yml
@@ -1,0 +1,17 @@
+---
+- name: Create a BareMetalPool resource
+  hosts: localhost
+  gather_facts: false
+
+  vars:
+    bare_metal_pool: "{{ ansible_eda.event.payload }}"
+    bare_metal_pool_name: "{{ bare_metal_pool.metadata.name }}"
+    bare_metal_pool_namespace: "{{ bare_metal_pool.metadata.namespace }}"
+    template_id: "{{ bare_metal_pool.metadata.annotations['osac.openshift.io/templateID'] }}"
+
+  tasks:
+    - name: Call the selected bare metal pool template
+      when: template_id != "" and template_id != "noop"
+      ansible.builtin.include_role:
+        name: "osac.templates.{{ template_id }}"
+        tasks_from: create
--- a/playbook_osac_delete_bare_metal_pool.yml
+++ b/playbook_osac_delete_bare_metal_pool.yml
@@ -1,0 +1,17 @@
+---
+- name: Delete a BareMetalPool resource
+  hosts: localhost
+  gather_facts: false
+
+  vars:
+    bare_metal_pool: "{{ ansible_eda.event.payload }}"
+    bare_metal_pool_name: "{{ bare_metal_pool.metadata.name }}"
+    bare_metal_pool_namespace: "{{ bare_metal_pool.metadata.namespace }}"
+    template_id: "{{ bare_metal_pool.metadata.annotations['osac.openshift.io/templateID'] }}"
+
+  tasks:
+    - name: Call the selected bare metal pool template for deletion
+      when: template_id != "" and template_id != "noop"
+      ansible.builtin.include_role:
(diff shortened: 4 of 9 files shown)


ajamias:
@tzumainn should we override the ci?

tzumainn:
> @tzumainn should we override the ci?

I think we can override the e2e-vmaas test; the others have been passing elsewhere, but a retest might be okay?

playbook_osac_create_bare_metal_pool.yml
```diff
@@ -0,0 +1,17 @@
...
+
+  vars:
+    bare_metal_pool: "{{ ansible_eda.event.payload }}"
+    bare_metal_pool_name: "{{ bare_metal_pool.metadata.name }}"
+    bare_metal_pool_namespace: "{{ bare_metal_pool.metadata.namespace }}"
+    template_id: "{{ bare_metal_pool.metadata.annotations['osac.openshift.io/templateID'] }}"
+
+  tasks:
+    - name: Call the selected bare metal pool template
+      when: template_id != "" and template_id != "noop"
```
DanNiESh, line 14: Is this check necessary? If the template_id is empty, the AAP job will fail to resolve the role then no role will be triggered anyway.

playbook_osac_delete_bare_metal_pool.yml
```diff
@@ -0,0 +1,17 @@
...
+
+  vars:
+    bare_metal_pool: "{{ ansible_eda.event.payload }}"
+    bare_metal_pool_name: "{{ bare_metal_pool.metadata.name }}"
+    bare_metal_pool_namespace: "{{ bare_metal_pool.metadata.namespace }}"
+    template_id: "{{ bare_metal_pool.metadata.annotations['osac.openshift.io/templateID'] }}"
+
+  tasks:
+    - name: Call the selected bare metal pool template for deletion
+      when: template_id != "" and template_id != "noop"
```
DanNiESh, line 14: same

playbook_osac_create_bare_metal_pool.yml
ajamias: It will fail but provisioning will keep being triggered from the baremetalpool because it's looking for a success.

collections/ansible_collections/osac/templates/roles/bm_private_network/defaults/main.yaml
```diff
@@ -0,0 +1,2 @@
+---
```
tzumainn, line 1: Can you leave a comment here mentioning that this is a temporary workaround, and what the future fix will be?

collections/ansible_collections/osac/templates/roles/bm_private_network/tasks/create.yaml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Extract network class from BareMetalPool status
+  ansible.builtin.set_fact:
+    network_class: "{{ bare_metal_pool.status.networkClass | default(default_network_class) }}"
+
```
tzumainn, line 5: Factually speaking, we do only support `openstack` right now so  maybe it makes sense to have an explicit check - but I guess the below task will essentially fail in the same way if someone uses a non-`openstack` network class?

tzumainn (commented):
Two very minor comments - looks good overall!

collections/ansible_collections/osac/templates/roles/bm_private_network/defaults/main.yaml
ajamias: yep

collections/ansible_collections/osac/templates/roles/bm_private_network/tasks/create.yaml
ajamias: Yes the plan is to default to openstack right now since `bare_metal_pool.status.networkClass` doesn't exist yet. And in a future update we can explicitly validate network classes.

tzumainn: I guess my question is whether you want to add that explicit validation now, if you're planning on adding it later?

ajamias: I plan to add it later

playbook_osac_create_bare_metal_pool.yml
tzumainn: I actually had the same thought as @DanNiESh, primarily because no other top level playbook has this check. But those other playbooks - cluster orders, virtual networks, etc - do not have optional templates, so the workflow will fail way back in CR creation if the template is not specified.

That being said, we do this `template_id` check in the controller as well, right? If so, I could see the argument for taking this check out.

ajamias: > But those other playbooks - cluster orders, virtual networks, etc - do not have optional templates, so the workflow will fail way back in CR creation if the template is not specified.

This check is needed because every BareMetalPool will trigger a template because there are templates like `agent_provisioning` where BareMetalPool doesn't need to trigger anything.

> That being said, we do this `template_id` check in the controller as well, right?

In the controller, template_id is set to "noop" by default. There is no check to prevent triggering the template.

tzumainn: > > But those other playbooks - cluster orders, virtual networks, etc - do not have optional templates, so the workflow will fail way back in CR creation if the template is not specified.
> 
> This check is needed because every BareMetalPool will trigger a template because there are templates like `agent_provisioning` where BareMetalPool doesn't need to trigger anything.
> 
> > That being said, we do this `template_id` check in the controller as well, right?
> 
> In the controller, template_id is set to "noop" by default. There is no check to prevent triggering the template.

The `agent_provisioning` profile has a template for host leases, and no template for bare metal pools, correct? Generally speaking, it would be correct for the operators to not trigger AAP if there is no template set. Can you create an issue for that?

ajamias: > The agent_provisioning profile has a template for host leases, and no template for bare metal pools, correct?

Yes that is correct

> Can you create an issue for that?

Yep here it is: https://github.com/osac-project/bare-metal-fulfillment-operator/issues/19

tzumainn (approved):
Thanks!

PR #314, opened 2026-05-21
Optionally attach node to a specified network after private network detach

DanNiESh:
After detaching a node from its private network, if `templateParameters.networkAfterDelete` is set, re-attach the node to
that network. If `templateParameters` is not provided or `networkAfterDelete` is not set, only the detach is performed.

### DIFF
--- a/collections/ansible_collections/osac/templates/roles/bm_host_private_network/tasks/delete.yaml
+++ b/collections/ansible_collections/osac/templates/roles/bm_host_private_network/tasks/delete.yaml
@@ -9,6 +9,10 @@
     msg: "bm_host_private_network requires networkClass 'openstack', got '{{ host_lease.spec.networkClass | default('') }}'"
   when: host_lease.spec.networkClass | default('') != "openstack"
 
+- name: Parse template parameters
+  ansible.builtin.set_fact:
+    template_params: "{{ host_lease.spec.templateParameters | default('{}') | from_json }}"
+
 - name: Detach node from private network
   ansible.builtin.include_role:
     name: massopencloud.esi.l2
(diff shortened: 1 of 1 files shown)


collections/ansible_collections/osac/templates/roles/bm_host_private_network/tasks/delete.yaml
```diff
@@ -17,6 +21,18 @@
...
     networks: []
 
+- name: Attach node to network after detach
+  ansible.builtin.include_role:
+    name: massopencloud.esi.l2
+    tasks_from: set_networks_for_node
+  vars:
+    node: "{{ host_lease.spec.externalHostID }}"
+    networks: "{{ [template_params.networkToAttach] }}"
+  when: template_params.networkToAttach | default('') | length > 0
```
tzumainn: Does `template_params.networkToAttach is defined` work as a check instead?

tzumainn (commented):
Can you rename `networkToAttach` to something like `networkAfterDelete`? Otherwise, the parameter sounds like it might be the network used in `create.yaml`.

collections/ansible_collections/osac/templates/roles/bm_host_private_network/tasks/delete.yaml
DanNiESh: Yes, changed to `template_params.networkAfterDelete is defined`

DanNiESh:
> Can you rename `networkToAttach` to something like `networkAfterDelete`? Otherwise, the parameter sounds like it might be the network used in `create.yaml`.

Done!

ajamias (approved):

tzumainn (approved):
Thanks!

PR #333, opened 2026-06-05
OSAC-1297: unbind agents before HostedCluster deletion to enable software reclaim

omer-vishlitzky:
## Summary

Add a pre-delete hook to the `ocp_ci_small` template that unbinds agents before deleting the HostedCluster, enabling the assisted-service software reclaim to run while the spoke API is still alive.

## Problem

When a ClusterOrder is deleted, the HostedCluster (spoke) is deleted first, then the agent controller tries to reclaim non-BMH agents. The software reclaim deploys a DaemonSet on the spoke to kexec the agent back into the discovery environment — but the spoke API is already gone, so reclaim fails silently. The agent gets permanently stuck in `unbinding-pending-user-action`.

This affects only non-BMH agents (libvirt VMs in CI). Bare metal with BMH/BMC is unaffected — Ironic handles the reboot via BMC independently.

## Fix

The `ocp_ci_small` template's pre-delete hook (previously a no-op) now calls `unbind_agents_for_reclaim`, which clears `spec.clusterDeploymentName` on each agent before the HostedCluster is deleted. This triggers the agent controller's unbind flow while the spoke is alive → `runReclaimAgent` deploys a DaemonSet on the spoke → the agent kexec-reboots into discovery → re-registers as `known-unbound`.

**Scope:** Only the `ocp_ci_small` template is affected. The generic `delete_hosted_cluster` flow and all other templates are untouched.

## Changes

- **New**: `manage_agents/tasks/unbind_agents_for_reclaim.yml` — clears `clusterDeploymentName`, waits for agents to reach `known-unbound`
- **New**: `ocp_ci_small/tasks/pre_delete.yaml` — calls `unbind_agents_for_reclaim`
- **Modified**: `ocp_ci_small/tasks/delete.yaml` — pre-delete hook points to `pre_delete.yaml` instead of `noop.yaml`
- **Modified**: `manage_agents/meta/argument_specs.yaml` — adds spec for new task

## Test plan

- [ ] Stress test on libvirt: verify agent returns to `known-unbound` after ClusterOrder deletion
- [ ] Verify non-`ocp_ci_small` templates are completely unaffected

Jira: [OSAC-1297](https://redhat.atlassian.net/browse/OSAC-1297)

## Summary by CodeRabbit

* **New Features**
  * Unbinds Agents during HostedCluster deletion for non-BMH deployments so Agent resources are cleared before cluster removal.
  * Adds a pre-delete task that triggers the agent-unbinding flow using the cluster order name.

* **Chores**
  * Default pre-delete hook now invokes the pre-delete task set so cleanup runs when no override is provided.

### DIFF
--- a/collections/ansible_collections/osac/service/roles/manage_agents/tasks/unbind_agents_for_reclaim.yml
+++ b/collections/ansible_collections/osac/service/roles/manage_agents/tasks/unbind_agents_for_reclaim.yml
@@ -44,11 +44,20 @@
         patch:
           - op: remove
             path: /spec/clusterDeploymentName
+      register: _unbind_patch_result
       loop: "{{ manage_agents_reclaim_list.resources }}"
       loop_control:
         loop_var: agent
         label: "Unbind agent {{ agent.metadata.name }}"
       failed_when: false
+
+    - name: Report unbind failures
+      ansible.builtin.debug:
+        msg: "Failed to unbind agent {{ item.agent.metadata.name }}: {{ item.msg | default('unknown error') }}"
+      loop: "{{ _unbind_patch_result.results | selectattr('failed', 'defined') | selectattr('failed') | list }}"
+      loop_control:
+        label: "{{ item.agent.metadata.name }}"
+      when: _unbind_patch_result.results | selectattr('failed', 'defined') | selectattr('failed') | list | length > 0
 
     - name: Wait for agents to return to available state
       kubernetes.core.k8s_info:
@@ -60,7 +69,7 @@
       register: manage_agents_reclaim_status
       until: >-
         manage_agents_reclaim_status.resources | length == 0
-        or manage_agents_reclaim_status.resources[0].status.debugInfo.state
+        or (manage_agents_reclaim_status.resources[0].status.debugInfo.state | default(''))
            in ['known-unbound', 'known', 'discovering-unbound']
       retries: 60
       delay: 10
--- a/collections/ansible_collections/osac/service/roles/manage_agents/tasks/unbind_agents_for_reclaim.yml
+++ b/collections/ansible_collections/osac/service/roles/manage_agents/tasks/unbind_agents_for_reclaim.yml
@@ -1,0 +1,71 @@
+---
+# Unbind agents BEFORE deleting the HostedCluster so the assisted-service software
+# reclaim can run while the spoke API is still alive.
+#
+# Problem: when a HostedCluster is deleted, the spoke API goes down. The assisted-service
+# agent controller then tries to reclaim non-BMH agents (libvirt VMs) by deploying a
+# DaemonSet on the spoke — but the spoke is gone, so reclaim fails silently. The agent
+# gets stuck in unbinding-pending-user-action permanently (it needs to kexec back into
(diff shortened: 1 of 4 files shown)


collections/ansible_collections/osac/service/roles/manage_agents/tasks/unbind_agents_for_reclaim.yml
```diff
@@ -0,0 +1,71 @@
...
+        kind: Agent
+        api_version: agent-install.openshift.io/v1beta1
+        namespace: "{{ default_agent_namespace }}"
+        field_selectors:
+          - "metadata.name={{ agent.metadata.name }}"
+      register: manage_agents_reclaim_status
+      until: >-
+        manage_agents_reclaim_status.resources | length == 0
+        or manage_agents_reclaim_status.resources[0].status.debugInfo.state
+           in ['known-unbound', 'known', 'discovering-unbound']
```
tzvatot, line 73: 🟡 **Unsafe nested field access in `until` expression**

`resources[0].status.debugInfo.state` has no null-safety. If any level is undefined (agent mid-transition, status not populated after kexec), Ansible treats the template error as `false` and retries - burning the full 10-minute timeout per agent with no useful work. The existing `wait_for_agents_to_be_removed.yml` in this same role uses `json_query` for safe field access. At minimum, guard with `| default('')`:

```yaml
until: >-
  manage_agents_reclaim_status.resources | length == 0
  or (manage_agents_reclaim_status.resources[0].status.debugInfo.state | default(''))
     in ['known-unbound', 'known', 'discovering-unbound']
```

collections/ansible_collections/osac/service/roles/manage_agents/tasks/unbind_agents_for_reclaim.yml
```diff
@@ -0,0 +1,71 @@
...
+        name: "{{ agent.metadata.name }}"
+        namespace: "{{ default_agent_namespace }}"
+        patch:
+          - op: remove
+            path: /spec/clusterDeploymentName
+      loop: "{{ manage_agents_reclaim_list.resources }}"
+      loop_control:
+        loop_var: agent
+        label: "Unbind agent {{ agent.metadata.name }}"
+      failed_when: false
```
tzvatot, line 52: 💡 **Register patch results for observability**

The `failed_when: false` here is the right call for a best-effort pre-delete hook (this replaces a `noop.yaml` - worst-case failure is identical to the old behavior). But registering the result and adding a debug task for failures would give operators visibility into when the optimization isn't working, without changing the failure behavior.

Re: CodeRabbit's suggestion to make this conditionally fail on RBAC/network errors - I'd push back on that. Making this hook fail on critical errors would block deletion entirely, trading a stuck-agent problem for a can't-delete-cluster problem. Best-effort is correct here.

tzvatot (changes requested):

tzvatot (approved):
## Re-review

| # | Previous Finding | Status |
|---|---------|--------|
| 1 | Unsafe nested field access in `until` expression | **FIXED** - `| default('')` guard added |
| 2 | Register patch results for observability | **FIXED** - result registered, debug task logs failures |

No new issues. Both findings addressed cleanly.

/lgtm

PR #354, opened 2026-06-18
NO-ISSUE: Fix race condition in bare metal template roles

### LINKED ISSUE #400: Ensure current bare metal template roles are idempotent
Right now, it seems as if some of the template roles implicitly assume that things run in a certain order. For example, during CaaS creation, the host lease template assumes that the private network created by the bare metal pool already exists; similarly, during deletion the bare metal pool deletes the private network without checking that the network has been detached from the hosts (resulting in stranded VIFs on the nodes).

We should ensure that these roles are idempotent and can theoretically run in any order; for example, the template that attaches the private network for the node could check to see if that network exists, and wait (and retry) if it does not.

ajamias:
Add network readiness checks so roles can run in any order: wait for the private network to exist in OpenStack before attaching a node (bm_host_private_network), and wait for all VIF ports to drain before deleting the network (bm_private_network). Also migrates role references from massopencloud.esi to osac.esi.

Assisted-by: Claude Code <noreply@anthropic.com>

Closes osac-project/issues#400

## Summary by CodeRabbit

# Release Notes

* **Bug Fixes**
  * Network attachment operations now include retry logic with explicit wait steps for network availability.
  * Network deletion now waits for all attached ports to be detached before proceeding.
  * Enhanced network configuration workflow stability with improved synchronization checks.

### DIFF
--- a/collections/ansible_collections/osac/templates/roles/bm_host_agent_deprovisioning/tasks/delete.yaml
+++ b/collections/ansible_collections/osac/templates/roles/bm_host_agent_deprovisioning/tasks/delete.yaml
@@ -11,7 +11,7 @@
 
 - name: Detach all networks and delete neutron ports
   ansible.builtin.include_role:
-    name: massopencloud.esi.l2
+    name: osac.esi.l2
     tasks_from: set_networks_for_node
   vars:
     node: "{{ bare_metal_instance.spec.externalHostID }}"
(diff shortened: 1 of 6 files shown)


ygalblum:
Code is OK. Just for house keeping as it will go into the git tree; I don't understand the title of the commit/pr. What in this change makes the role idempotent?

ajamias:
> What in this change makes the role idempotent?

@ygalblum My mistake, I updated the title to state the change as a race condition fix

ygalblum (approved):
Thanks for clarifying

/approve
/lgtm

adriengentil (approved):

PR #367, opened 2026-06-23
OSAC-1698: Revert VM network spec to pod: {} with bridge interface

ori-amizur:
Replace multus default:true NAD reference with pod: {} in the VM spec. The explicit Multus reference causes failures on clusters with the OVN-K enable-preconfigured-udn-addresses feature gate enabled, because OVN-K expects the default-network annotation to point to openshift-ovn-kubernetes, not a user namespace. With pod: {} the CUDN is still the primary network via the namespace label k8s.ovn.org/primary-user-defined-network, and the playbook validation still prevents VMs without a subnet reference.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated default network attachment configuration in VM templates to properly resolve network interfaces for both Linux and Windows guest operating systems using pod-scoped network configuration.

### DIFF
--- a/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/tasks/create_build_spec.yaml
+++ b/collections/ansible_collections/osac/templates/roles/ocp_virt_vm/tasks/create_build_spec.yaml
@@ -38,9 +38,7 @@
               spinlocks: 8191
       networks:
         - name: default
-          multus:
-            networkName: "{{ compute_instance_target_namespace }}/{{ vm_cudn_nad_name }}"
-            default: true
+          pod: {}
       volumes: "{{ root_disk_volume }}"
   when: guest_os_family != 'windows'
 
@@ -93,9 +91,7 @@
               spinlocks: 8192
       networks:
         - name: default
-          multus:
-            networkName: "{{ compute_instance_target_namespace }}/{{ vm_cudn_nad_name }}"
-            default: true
+          pod: {}
       volumes: "{{ root_disk_volume }}"
   when: guest_os_family == 'windows'
 


collections/ansible_collections/osac/templates/roles/ocp_virt_vm/tasks/create_build_spec.yaml
```diff
@@ -38,9 +38,7 @@
               spinlocks: 8191
       networks:
         - name: default
-          multus:
```
sk-ilya, line 41: I'm dealing with a VM failure caused by the network configuration, and I suspect this reverts to a config that was already incorrect.

I think what's needed is `binding: {name: l2bridge}` + `pod: {}` - that fixed the problem with CUDN  for me

achuzhoy:
I tested this PR on my setup - was able to successfully launch a VM with osac.

achuzhoy (approved):
Used this PR on the same setup where I saw the failure and was able to spawn a VM with no issues.

ygalblum:
I understand why this change solves the issue. But, is it the correct solution?
@ori-amizur I see you made the pervious change. Why was it wrong? Also, should the VMs always be connected to the `pod` network? That might give them unintended network access. No?

ori-amizur:
Good questions, let me clarify both.

**Why was the `multus: {default: true}` approach wrong?**

The explicit Multus NAD reference worked on our test cluster but broke on clusters with the OVN-K `enable-preconfigured-udn-addresses` feature gate enabled. That gate changes how OVN-K validates the `default-network` annotation — it expects it to point to `openshift-ovn-kubernetes`, not a user namespace. So the approach turned out to be environment-dependent, which makes it fragile.

**Should VMs always be connected to the `pod` network? Doesn't that give them unintended network access?**

In our setup, `pod: {}` does NOT give access to the default cluster network. Here's why:

The subnet namespace is created with the label `k8s.ovn.org/primary-user-defined-network: ""` (in `cudn_net/tasks/create_subnet.yaml`), and the CUDN is created with `role: Primary`. This makes OVN-K replace the default pod network with the CUDN for all pods in that namespace. So when KubeVirt sees `pod: {}`, it connects the VM to the namespace's primary network — which IS the CUDN, not the default cluster network.

The VM gets an IP from the CUDN's CIDR range and has no path to the default cluster network. The `pod: {}` is just an indirection to whatever the namespace's primary network is — not a fallback.

So reverting to `pod: {}` is correct: it's safe (no unintended access), portable (works across OVN-K feature gate configurations), and the CUDN is always used as long as the namespace label is in place.

danmanor (approved):

ygalblum:
@ori-amizur thanks for the explanation. I see that the PR was already merged

PR #374, opened 2026-06-24
OSAC-1693: Add importAgentsEnabled to config-as-code secret template

trewest:
Broaden the config-as-code-secret gate to also fire when importAgentsEnabled is true, and conditionally include OSAC_IMPORT_AGENTS_ENABLED in the secret. This allows the osac-installer parent chart to control import agent enablement via Helm value passthrough (aap.configAsCode.importAgentsEnabled) without needing a separate template that conflicts with this one.


eliorerz (approved):
LGTM

PR #399, opened 2026-07-07
OSAC-1962: add AGENTS.md for AI agent onboarding

amej:
## Summary
- Adds AGENTS.md documenting the template-role pattern, meta/osac.yaml structure, Python version discrepancy (3.13 local / 3.12 container), and ansible-test integration testing conventions

## Jira
https://redhat.atlassian.net/browse/OSAC-1962

## Test plan
- [x] `make lint` (ansible-lint via uv) passes: 0 failures, 32 pre-existing warnings
- [x] Every role/path/version referenced in AGENTS.md verified against the live repo

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -4,12 +4,12 @@ Guide for AI agents working on osac-aap — Ansible Automation Platform roles an
 
 ## Overview
 
-osac-aap contains Ansible collections that provision and manage OSAC infrastructure:
-- **Network backends**: CUDN, Netris, OpenStack, MetalLB
-- **Compute backends**: OpenShift Virtualization (KubeVirt)
-- **Cluster provisioning**: OpenShift 4.17, 4.20 templates
-- **Storage backends**: VAST
-- **Bare metal provisioning**: ESI, Metal3, NICo
+osac-aap contains Ansible collections that provision and manage OSAC infrastructure. Backends are added over time — the lists below are examples, not exhaustive:
+- **Network backends**: e.g. CUDN, Netris, OpenStack, MetalLB
+- **Compute backends**: e.g. OpenShift Virtualization (KubeVirt)
+- **Cluster provisioning**: e.g. OpenShift 4.17, 4.20 templates
+- **Storage backends**: e.g. VAST
+- **Bare metal provisioning**: e.g. OpenStack Ironic (via the ESI collection), Metal3, NICo
 
 Integration flow: fulfillment-service (API) → osac-operator (CRs) → AAP job templates → osac-aap playbooks → template roles
 
@@ -22,6 +22,8 @@ Install dependencies:
 uv sync --all-groups
 source .venv/bin/activate  # or prefix commands with `uv run`
 ```
+
+`uv sync` creates or updates `.venv` in place to match `pyproject.toml`/`uv.lock` — re-run it after dependency changes instead of recreating the venv manually.
 
 ### Key Configuration Files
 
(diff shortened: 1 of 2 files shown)


amej:
@coderabbitai rate limit

AGENTS.md
```diff
@@ -0,0 +1,512 @@
...
+Guide for AI agents working on osac-aap — Ansible Automation Platform roles and playbooks for OSAC infrastructure provisioning.
+
+## Overview
+
+osac-aap contains Ansible collections that provision and manage OSAC infrastructure:
+- **Network backends**: CUDN, Netris, OpenStack, MetalLB
+- **Compute backends**: OpenShift Virtualization (KubeVirt)
+- **Cluster provisioning**: OpenShift 4.17, 4.20 templates
+- **Storage backends**: VAST
+- **Bare metal provisioning**: ESI, Metal3, NICo
```
ygalblum: I'm not sure if we should list the options as it will require updating the file when adding functionality. Should we at the least say that these are examples and that others will follow?

AGENTS.md
```diff
@@ -0,0 +1,512 @@
...
+
+Integration flow: fulfillment-service (API) → osac-operator (CRs) → AAP job templates → osac-aap playbooks → template roles
+
+## Development Environment
+
+### Setup
+
+Install dependencies:
+```bash
+uv sync --all-groups
```
ygalblum, line 22: Does this update the `.venv`?

AGENTS.md
```diff
@@ -0,0 +1,512 @@
...
+├── execution-environment/          # Container build definitions
+├── charts/aap/                     # Helm charts for AAP deployment
+└── samples/                        # Example payloads
+```
+
+## Template Role Pattern
+
+Each template role in `osac.templates` declares capabilities in `meta/osac.yaml`:
+
+```yaml
```
ygalblum: I don't understand this snippet. IIUC it has an example for networking and then a bunch of comments about cluster. Then the specific examples are below. Is it beneficial?

AGENTS.md
```diff
@@ -0,0 +1,512 @@
...
+Assisted-by: Claude Code <noreply@anthropic.com>
+```
+
+Note: Use `Assisted-by`, not `Co-Authored-By` (Red Hat attribution standard).
+
+### Fork-Based Workflow
+
+- Push to `fork` remote, not `origin`
+- PRs go from `fork/<branch>` to `origin/main`
+- Always include Jira ticket key in PR title (e.g., "OSAC-12345: add VAST storage backend")
```
ygalblum: Here you say always use the JIRA ticket but before you allow not to have it

AGENTS.md
ygalblum, line 1: I wonder if this file is not too long and whether it can be split into skills. From what I can tell, specific agents will not need all this information as they are supposed to be more dedicated.

AGENTS.md
tzumainn: I would also encourage the use of `OpenStack` instead of `ESI`

AGENTS.md
```diff
@@ -0,0 +1,512 @@
...
+
+## Common Fix Locations
+
+Use this table to go directly to the right file for common patterns:
+
+| Pattern | Location |
+|---------|----------|
+| Add new network backend | `collections/ansible_collections/osac/templates/roles/<backend>/` + `meta/osac.yaml` |
+| Add new compute backend | `collections/ansible_collections/osac/templates/roles/<backend>/` + `meta/osac.yaml` |
+| Add new cluster template | `collections/ansible_collections/osac/templates/roles/<template>/` + `meta/osac.yaml` |
```
tzumainn, line 374: It may make sense to add bare metal templates here as well!

AGENTS.md
amej: Good point — updated to mark these as examples (`e.g.`) with a note that backends are added over time, so the list doesn't need to be kept exhaustive.

AGENTS.md
amej: Yes — added a line clarifying that `uv sync` creates/updates `.venv` in place to match `pyproject.toml`/`uv.lock`.

AGENTS.md
amej: Agreed it wasn't pulling its weight — simplified this into a one-line pointer to the per-type `meta/osac.yaml` breakdown further down, which already documents each template_type (including the cluster exception) in full.

AGENTS.md
amej: Fixed — added `NO-ISSUE: description` as an explicit valid format in Commit Messages (it's used in this repo's history), and updated the PR title guidance to say the ticket key is included when one applies, `NO-ISSUE` otherwise, matching the commit convention.

AGENTS.md
amej: Fair concern — I left the single-file structure as-is for now rather than unilaterally splitting it, since that's a bigger structural call. Trimmed some redundancy in this pass (e.g. the Template Role Pattern intro snippet). If we want to split into skills for dedicated agents, happy to follow up with that as a separate change — let me know if you'd like to see a proposal.

AGENTS.md
amej: Updated — the bare metal line now reads "OpenStack Ironic (via the ESI collection), Metal3, NICo" to lead with the underlying tech, consistent with how the README describes it.

AGENTS.md
amej: Added a "Add new bare metal template" row to the Common Fix Locations table, pointing to the same role-directory + meta/osac.yaml pattern as the other backend types.

amej (commented):
I have addressed the reviewer's comments. Updated the CLAUDE.md as per the https://code.claude.com/docs/en/memory#agents-md

ygalblum (approved):
/approve
/lgtm

PR #402, opened 2026-07-07
OSAC-1699: Revert QoS policies created with wrong type for block storage tiers

wgordon17:
Reverts osac-project/osac-aap#393

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated QoS policy creation to support VAST 5.4.x constraints by only using supported policy types and excluding block-based QoS options.
  * Block storage class setup no longer adds QoS policy settings, reducing invalid configuration for block tiers.

* **Tests**
  * Adjusted integration test data and expectations to match the updated storage tier and QoS behavior.


wgordon17:
This PR moves the state back to a _previously_ known working state (it's a straightforward PR/commit revert). I'm proceeding with a self-merge

PR #408, opened 2026-07-08
OSAC-2286: add instance groups, API token hook, config-as-code-manifest

omer-vishlitzky:
## Summary

- Add instance group templates (cluster/network/storage/publish-templates IGs) with ConfigMap+Secret stubs when disabled
- Add API token post-install hook (creates osac-aap-api-token Secret)
- Add config-as-code-manifest Secret template
- Remove `optional: true` from bootstrap config-as-code-ig secretRef (secret is now chart-managed)

Part of [OSAC-1928](https://redhat.atlassian.net/browse/OSAC-1928) umbrella cleanup. Tested atomically via [osac-installer PR #404](https://github.com/osac-project/osac-installer/pull/404).

## Test plan

- [ ] `helm template` renders correctly
- [ ] e2e passes via installer PR #404

## Summary by CodeRabbit

* **New Features**
  * Added optional Config-as-Code manifest Secret creation and introduced new chart defaults.
  * Added post-install/post-upgrade hooks to create an AAP API token and publish templates automatically.
  * Expanded instance group support with conditionally rendered ConfigMaps/Secrets for cluster, network, storage, and publish templates, including hosted base domain detection.
* **Bug Fixes**
  * Adjusted bootstrap startup probing to preserve curl error output during readiness checks.
  * Updated bootstrap environment sourcing to require the configured Secret (removed the optional fallback).


charts/aap/templates/hooks/publish-templates.yaml
```diff
@@ -0,0 +1,151 @@
...
+        command:
+          - /bin/bash
+          - -euo
+          - pipefail
+          - -c
+          - |
+            echo "Waiting for fulfillment internal API..."
+            for i in $(seq 1 60); do
+              echo "Attempt ${i}: checking http://fulfillment-ingress-proxy-metrics:8002/healthz"
+              if curl -sf http://fulfillment-ingress-proxy-metrics:8002/healthz; then
```
jhernand: This is assuming that the fulfillment service is running in the same cluster than the operator, which is conceptually wrong. In addition, even if it was in the same cluster, it is using a the heath URL of the metrics endpoint of Envoy. That is an internal implementation detail that can change without notice. This should use the URL of the API.

jhernand (changes requested):

jhernand (approved):

PR #410, opened 2026-07-10
NO-ISSUE: creates test HyperConverged CR via v1 to fix CI flake

wgordon17:
## Summary
- The integration test setup creates the test HyperConverged CR through v1beta1, while the ocp_virt_vm role's GPU passthrough task (and the test's own verification step) read and patch it through v1
- Where the two API versions' schemas diverge for spec.featureGates (object under v1beta1, array under v1, no conversion webhook between them), this cross-version mismatch causes any later write to fail full-object CEL validation with a confusing type-mismatch error surfaced as a uniqueness rule violation
- Creating the CR through v1 for consistency with the rest of the flow fixes compute_instance_with_gpu_create:baseline; reproduced and verified locally with a kind cluster before and after the change

## Summary by CodeRabbit

- **Tests**
  - Updated the integration test environment to use the current HyperConverged API version.
  - Added guidance documenting API version selection, defaulting, and validation expectations to improve test consistency and reliability.

### DIFF
--- a/tests/integration/setup_test_env.sh
+++ b/tests/integration/setup_test_env.sh
@@ -93,8 +93,13 @@ kubectl create namespace openshift-cnv || true
 
 # 5. Create minimal HyperConverged CR for GPU passthrough tests
 echo "Creating HyperConverged CR for GPU passthrough tests..."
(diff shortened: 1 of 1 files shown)


akshaynadkarni (approved):
LGTM

PR #416, opened 2026-07-14
OSAC-2185: Gate chart publish on image build success

minmzzhang:
## Summary

`publish-charts.yaml` and `execution-environment.yml` currently trigger independently off the same `v*` tag push, with no dependency between them — a chart can be marked "published" even if the matching execution-environment image never got built. This exact class of gap caused the [OSAC-2174](https://redhat.atlassian.net/browse/OSAC-2174) production incident in a sibling repo (chart referenced an image that was never built, resulting in `ImagePullBackOff`).

This PR changes `publish-charts.yaml` to trigger via `workflow_run` on `Build execution environment` completing, gated by a new `guard` job:

- Skips (neutral, not a false green) if the completion isn't from a tag push (e.g. PR builds, pushes to `main`).
- **Fails loudly (red)** if the image build for the tag did not succeed.
- Only lets the chart-publish job proceed if the image build for that *same tag* succeeded.

The single `publish-aap-chart` job now `needs: guard`, pins its checkout to `needs.guard.outputs.sha`, and uses `needs.guard.outputs.tag` instead of `$GITHUB_REF_NAME` (which doesn't reliably resolve to the tag under a `workflow_run` event) for both version extraction and the inline `gh release create` step.

## Testing

Live end-to-end verified on a personal fork, not just static analysis:
- Pushed a `v*` test tag → image build ran → succeeded → `publish-charts` triggered afterward (not in parallel) → chart published → GitHub Release created. Full run: green.
- Confirmed the negative path: a `push` to `main` (non-tag) completing the image workflow correctly results in `publish-charts` being **skipped**, not a false green.
- Validated with `actionlint` (0 errors) on the final workflow file.

## Acceptance criteria ([OSAC-2185](https://redhat.atlassian.net/browse/OSAC-2185))

- [x] Chart publish cannot succeed (or create a GitHub Release) unless the matching image was confirmed to exist for that tag.
- [x] A simulated image-build failure on a tag push causes `publish-charts` to fail loudly rather than complete green.

Jira: https://redhat.atlassian.net/browse/OSAC-2185

## Summary by CodeRabbit

* **Chores**
  * Updated chart publishing to trigger only after the build execution workflow finishes successfully, rather than on direct version tag pushes.
  * Added validation to ensure publishing is based on a proper version tag from the expected branch and that the upstream workflow concluded successfully.
  * Added an extra verification step to confirm the guarded tag still resolves to the expected commit before creating/updating the GitHub Release, with a dedicated script performing the check.

### DIFF
--- a/.github/scripts/verify-tag-matches-sha.sh
+++ b/.github/scripts/verify-tag-matches-sha.sh
@@ -7,7 +7,8 @@
 # Required env vars: GH_TOKEN, REPO, TAG, GUARDED_SHA
 set -euo pipefail
 
-read -r current_type current_sha <<< "$(gh api "repos/${REPO}/git/refs/tags/${TAG}" --jq '[.object.type, .object.sha] | @tsv')"
+ref_json="$(gh api "repos/${REPO}/git/ref/tags/${TAG}" --jq '[.object.type, .object.sha] | @tsv')"
+read -r current_type current_sha <<< "$ref_json"
 if [[ "$current_type" == "tag" ]]; then
   # Annotated tag: the ref's object.sha is the tag object, not the commit - peel it.
   current_sha="$(gh api "repos/${REPO}/git/tags/${current_sha}" --jq .object.sha)"
--- a/.github/scripts/verify-tag-matches-sha.sh
+++ b/.github/scripts/verify-tag-matches-sha.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+# Verifies that a git tag still resolves to the guarded commit SHA that was
+# validated by this workflow's guard job, protecting against a force-push/
+# retag race between the image build and any subsequent chart-publish or
+# release step.
+#
+# Required env vars: GH_TOKEN, REPO, TAG, GUARDED_SHA
+set -euo pipefail
+
+read -r current_type current_sha <<< "$(gh api "repos/${REPO}/git/refs/tags/${TAG}" --jq '[.object.type, .object.sha] | @tsv')"
+if [[ "$current_type" == "tag" ]]; then
+  # Annotated tag: the ref's object.sha is the tag object, not the commit - peel it.
+  current_sha="$(gh api "repos/${REPO}/git/tags/${current_sha}" --jq .object.sha)"
+fi
+
+if [[ "$current_sha" != "$GUARDED_SHA" ]]; then
+  echo "::error::Tag '${TAG}' now points at ${current_sha}, not the guarded commit ${GUARDED_SHA} (force-pushed?). Refusing to proceed."
+  exit 1
+fi
+
+echo "Tag '${TAG}' still points at the guarded commit ${GUARDED_SHA}."
--- a/.github/workflows/publish-charts.yaml
+++ b/.github/workflows/publish-charts.yaml
@@ -13,6 +13,8 @@ jobs:
   guard:
     name: Verify image build succeeded
     runs-on: ubuntu-latest
+    # Read-only job: only inspects workflow_run event metadata, no repo/API access needed.
+    permissions: {}
     if: >
       github.event.workflow_run.event == 'push' &&
       startsWith(github.event.workflow_run.head_branch, 'v')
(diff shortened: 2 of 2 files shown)


minmzzhang:
Pushed a follow-up commit addressing the CodeRabbit review:

- `guard` job now runs with `permissions: {}` (was inheriting default token scope)
- `workflow_run.head_branch`/`conclusion` are now passed through `env:` instead of being spliced directly into the shell script (script/template-injection hardening)
- The tag is now validated against a proper semver grammar in `guard`, not just `startsWith(..., 'v')`
- `persist-credentials: false` and a SHA-pinned `actions/checkout` on the touched checkout step(s)
- Added a step that re-verifies the tag still resolves to the guarded commit immediately before creating the GitHub Release, to guard against a force-push/retag race between the image build and the release

Intentionally not doing in this PR: Sigstore/cosign signing of the published chart artifacts. That's a separate, larger effort (registry/verification infra, signing key strategy) that's outside [OSAC-2185](https://redhat.atlassian.net/browse/OSAC-2185)'s scope (gating chart publish on image build success). Happy to track it as its own follow-up ticket if wanted.

minmzzhang:
Pushed a follow-up commit addressing round 2 of the CodeRabbit review:

- Extracted the tag-vs-guarded-commit verification into a shared `.github/scripts/verify-tag-matches-sha.sh` instead of duplicating the bash logic at each checkpoint (also flagged as a concern independently - thanks for catching that)
- The script now dereferences annotated tags before comparing SHAs (the ref lookup returns the tag-object SHA, not the commit SHA, for annotated tags)
- The verification now runs both right after checkout (before packaging/pushing the chart) and again immediately before creating the release, closing the gap where a moved tag would only be caught after an artifact was already uploaded
- Added `--verify-tag` to `gh release create` as an extra safety net
- Fixed the SemVer regex to reject leading zeros (e.g. `v01.2.3`, `v1.2.3-01`)

One finding I did **not** apply: the claim that `git/refs/tags/...` is "the wrong endpoint and will fail" - I tested both `git/refs/tags/...` and `git/ref/tags/...` directly against the live GitHub API and they return identical results, so this doesn't hold up. Kept the original endpoint.

minmzzhang:
Pushed a follow-up commit applying the same fix pattern used for this identical workflow across the [OSAC-2185](https://redhat.atlassian.net/browse/OSAC-2185) PRs:

- Removed the `+build` metadata group from the SemVer regex - Docker/OCI tags can't contain `+`, so a tag like `v1.2.3+build.1` would have passed validation but produced an unusable image reference downstream
- Routed the remaining direct `${{ secrets.GITHUB_TOKEN }}`/`${{ github.actor }}`/`${{ github.repository }}` interpolations (in the `Log in to GHCR` / `Create GitHub Release` steps) through `env:` instead of splicing them into the shell - consistent with the pattern already applied to the other steps in this file

minmzzhang:
Pushed a follow-up commit addressing round 4 of the CodeRabbit review:

- Fixed the `gh api`-into-`read` failure masking in `verify-tag-matches-sha.sh` - the API response is now captured in a variable first, so a failed call surfaces its real error under `set -e` instead of falling through to a generic tag-mismatch message
- Switched to the officially documented singular `git/ref/{ref}` endpoint for retrieval - confirmed against GitHub's actual REST API docs that the plural `git/refs/{ref}` form is only documented for `PATCH`/`DELETE`, not `GET` (both happen to return identical data today, but only the documented form is guaranteed to keep doing so)
- Pinned `azure/setup-helm` to a full commit SHA - it was still on the mutable `v5` tag even though `actions/checkout` was already pinned
- Routed the workflow-level `REGISTRY` env var through `"$REGISTRY"` directly in the `Log in to GHCR` step instead of `${{ env.REGISTRY }}`, consistent with the rest of the file

rccrdpccl:
@minmzzhang could this be a shared job instead?

minmzzhang:
> @minmzzhang could this be a shared job instead?

Very good suggestion. I kept the change local because currently each repo still needs its own `workflow_run` trigger which are different upstream workflow names, and the publish jobs diverge per chart/path/release behavior. Extracting the shared `guard` into a reusable workflow fits better as a follow-up once we look at the overall CI structure. What do you think?

rccrdpccl:
/lgtm
/approve

as discussed offline with @minmzzhang we should extract common parts to a common workflow later

PR #428, opened 2026-07-19
OSAC-2535: Remove public-ip references from osac-aap

ori-amizur:
Now that external-ip resources are merged, remove all remaining public-ip playbooks, task files, argument specs, AAP job templates, and documentation references. The agentless_net collection is left untouched as it refers to literal public IP addresses, not the OSAC resource type.


ygalblum:
@ori-amizur is there a reason why this PR is in draft?

danmanor (approved):

PR #441, opened 2026-07-23
[WIP] Allow per-resource-class route metric override for NMStateConfig

eliorerz:
## Summary
- `netris.steps.cluster_infra`'s create.yaml always applies `osac.service.nmstate_config` with that role's hardcoded route-metric defaults (mgmt=100, vpc=50 -- vpc wins), which is correct for the original Netris-hosted-control-plane design but wrong for hybrid designs where the control plane lives behind the mgmt network instead.
- Threads `nmstate_config_mgmt_route_metric`/`nmstate_config_vpc_route_metric` from the resource class's own optional `mgmt_route_metric`/`vpc_route_metric` fields via `| default(omit)`, so a resource class needing mgmt to win can set these in `netris_resource_class_map` without affecting any other resource class.
- Fully backward compatible: no existing resource class currently sets these fields, so this is a no-op for all of them -- they keep falling through to the role's own defaults, unchanged from today.

## Context
Found via osac-test-infra's `e2e-caas-netris-full-install` CI (run 29968016281), a new hybrid flow that combines a fast cluster-tool-booted control plane (living behind `br-mgmt`) with real Netris bare-metal discovery/provisioning for workers. Confirmed live that leaving `mgmt_route_metric`/`vpc_route_metric` unset applies the role's defaults byte-for-byte -- the joining worker's `nmconnection` files showed exactly `metric=100`/`metric=50` -- sending the worker's own outbound/API-join traffic to the Netris fabric instead of back through mgmt to the real (mgmt-hosted) control plane.

The companion osac-test-infra change registers its "ci-worker" resource class with `mgmt_route_metric: 40, vpc_route_metric: 50` once this PR is available, on branch `OSAC-2801-caas-netris-hybrid-install`.

## Test plan
- [ ] `ansible-lint` clean (verified locally -- zero new warnings vs. main, confirmed via git stash diff)
- [ ] Validate against osac-test-infra's `e2e-caas-netris-full-install` CI once this lands (or is tested via a branch override), confirming the previously-stuck worker join now completes with `ens4`/mgmt winning the default route
- [ ] Confirm no regression to existing resource classes (none set these new fields, so behavior should be identical)

## Summary by CodeRabbit

* **New Features**
  * Added support for optional management and VPC routing metric overrides during network configuration, applied per resource class with sensible defaults when unset.
  * Overrides are honored for both existing cluster agents and newly added agents, preserving prior precedence for VPC metrics.
  * Introduced a new `netris_hybrid_control_plane` setting (from `NETRIS_HYBRID_CONTROL_PLANE`) to tailor DNS/API behavior in hybrid deployments.
* **Bug Fixes**
  * Adjusted API DNAT creation and deletion logic so API-side DNAT is skipped in hybrid mode.

### DIFF
--- a/collections/ansible_collections/netris/steps/roles/cluster_infra/tasks/create.yaml
+++ b/collections/ansible_collections/netris/steps/roles/cluster_infra/tasks/create.yaml
@@ -225,10 +225,19 @@
 # mgmt_route_metric/vpc_route_metric are optional per-resource-class
 # overrides (netris_resource_class_map[<class>].mgmt_route_metric /
 # .vpc_route_metric). Resource classes that don't set them fall through to
-# osac.service.nmstate_config's own defaults (mgmt=100, vpc=50 -- vpc wins),
-# which is what every existing resource class already relies on implicitly.
-# Needed for hybrid control-plane designs where mgmt must win instead (the
-# vpc/Netris-fabric network isn't the production network there).
+# the literal defaults below (100/50, mirroring osac.service.nmstate_config's
+# own defaults/main.yaml -- vpc wins), which is what every existing resource
+# class already relies on implicitly. Needed for hybrid control-plane
+# designs where mgmt must win instead (the vpc/Netris-fabric network isn't
+# the production network there).
+#
+# These use an explicit literal default, NOT `| default(omit)`: confirmed
+# empirically (ansible-core 2.18) that `default(omit)` does not propagate
+# correctly through include_role's `vars:` when the source value is
+# undefined -- it sets the variable to a literal
+# "__omit_place_holder__..." string instead of leaving it unset, which
+# would have broken NMState application for every resource class that
+# doesn't set these fields (i.e. every one today).
 - name: Create NMStateConfig CRs for all cluster agents
   ansible.builtin.include_role:
     name: osac.service.nmstate_config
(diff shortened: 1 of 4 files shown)


eliorerz:
Thanks for the review -- verified both findings against the current code:

**Finding 2 (`default(omit)` in include_role vars) -- confirmed and fixed.** Reproduced with a minimal role/playbook against ansible-core 2.18: `myvar: "{{ x | default(omit) }}"` passed through `include_role`'s `vars:` does NOT correctly omit the variable when `x` is undefined -- it sets `myvar` to a literal `__omit_place_holder__<hash>` string instead of falling through to the included role's own default. `omit` only reliably works as a direct module-parameter value or inside `set_fact`, not as an `include_role` vars value. As written, this would have broken NMState route-metric application for every resource class that doesn't set `mgmt_route_metric`/`vpc_route_metric` -- i.e. every one today. Fixed in 548c8b7 by using explicit literal defaults (100/50, matching `osac.service.nmstate_config`'s own `defaults/main.yaml`) instead, verified against both the undefined and defined-override cases.

**Finding 1 (resolve metrics per resource-class group, not `cluster_infra_node_requests[0]`) -- valid but out of scope for this PR, skipping.** Confirmed this single-resource-class assumption already exists in the current code for `_netris_resource_class`/`_server_mgmt_interface`/`_server_vpc_interfaces`/`_netris_server_cluster_template_id` (all derived from `cluster_infra_node_requests[0]` before this PR touched anything) -- this PR's two new fields just inherit that existing limitation, they don't introduce a new one. Restructuring the whole role to group by resource class and loop the downstream Netris server-cluster/NAT/SNAT/NMState logic per group is a real improvement but a much larger, separate change. Happy to file a follow-up issue for it if useful.

PR #449, opened 2026-07-26
OSAC-2915: Add unit tests for template parameter pydantic validation

tzvatot:
## [OSAC-2915](https://redhat.atlassian.net/browse/OSAC-2915): Add unit tests for template parameter pydantic validation

**Jira:** [OSAC-2915](https://redhat.atlassian.net/browse/OSAC-2915)
**Related:** [OSAC-2816](https://redhat.atlassian.net/browse/OSAC-2816) (list/dict fix, merged in #430)

### Summary

Adds the first Python unit test suite for `find_template_roles.py` pydantic models. These tests parse real template role YAML through the pydantic models and verify `ProtobufAnyValue` serialization across the full type matrix, so that schema/data mismatches are caught by CI instead of silently dropping templates (as happened with [OSAC-2816](https://redhat.atlassian.net/browse/OSAC-2816)).

### Changes

- **`pyproject.toml`**: Added `pytest>=8.0.0` and `pytest-cov>=6.0.0` to dev dependency group
- **`tests/unit/`**: New unit test directory structure with conftest providing `roles_dir` fixture
- **`tests/unit/plugins/filter/test_find_template_roles.py`**: 55 unit tests covering:
  - `TemplateParameterDefinition` defaults for all types (str, int, float, bool, list, dict, None)
  - `TemplateParameter.from_definition()` type mapping and `ProtobufAnyValue` wrapping
  - `ProtobufAnyValue` serialization with `@type` alias
  - Real `osac.yaml` parsing for all template roles in the repo
  - `TypeMapping` completeness for Ansible string keys and Python built-in types
  - `Metadata` parsing for all `template_type` variants

### Testing

- **Unit tests:** 55 tests, all passing (0.18s)
- **Integration tests:** N/A (pure data model tests, no component interactions)
- **Run:** `uv run pytest tests/unit/ -v`

### Follow-up

- Add a `pytest` CI workflow step so these tests run on PRs (not in scope for this story)

### Acceptance Criteria

- [ ] AC-1: Unit tests that parse real template role YAML through pydantic models
- [ ] AC-2: Cover the full type matrix: str, int, float, bool, list, dict defaults
- [ ] AC-3: Verify all discovered templates produce valid ProtobufAnyValue serialization

## Summary by CodeRabbit

* **Tests**
  * Added unit and integration-style coverage for template parameter default handling across scalars, `list`/`dict` (including empty), and `None`.
  * Verified correct default-to-Type mapping, protobuf `Any` serialization (including `@type` presence), and proper validation errors for invalid defaults.
  * Added regression checks to ensure `list`/`dict` defaults are accepted and preserved, plus repository-backed OSAC role metadata assertions.

* **Chores**
  * Updated the development dependency configuration to include `pytest` and `pytest-cov`, formatted for clarity.


tests/conftest.py
```diff
@@ -0,0 +1,15 @@
+import sys
+from pathlib import Path
+
+sys.path.insert(
```
alosadagrande: Nit: `sys.path.insert` is fragile — if someone moves the filter plugin directory, this breaks silently. An alternative would be to set `PYTHONPATH` in `[tool.pytest.ini_options]` in `pyproject.toml`:

```toml
[tool.pytest.ini_options]
pythonpath = ["collections/ansible_collections/osac/service/plugins/filter"]
```

Not a blocker, just a suggestion for a cleaner approach.

pyproject.toml
```diff
@@ -22,7 +22,12 @@ dependencies = [
 ]
 
 [dependency-groups]
-development = ["ansible-lint>=25.2.1", "antsibull-changelog>=0.33.0"]
+development = [
+  "ansible-lint>=25.2.1",
+  "antsibull-changelog>=0.33.0",
```
alosadagrande, line 27: Minor inconsistency: the existing dev deps use `>=` (e.g., `ansible-lint>=25.2.1`) while pytest and pytest-cov are pinned with `==`. I'm fine either way, but it's worth being intentional — if pinning is preferred, consider pinning the others too for consistency, or use `>=` here to match the existing style.

tests/unit/plugins/filter/test_find_template_roles.py
```diff
@@ -0,0 +1,288 @@
...
+
+    def test_cluster_with_parameters(self, roles_dir):
+        metadata = _load_metadata(roles_dir, "ocp_4_20_ai_maas")
+        assert len(metadata.parameters) > 0
+        param_names = [p.name for p in metadata.parameters]
+        assert "hardware_profiles" in param_names
+        assert "external_models" in param_names
+
+    def test_validate_default_rejects_unsupported_type(self):
+        with pytest.raises(pydantic.ValidationError):
```
alosadagrande, line 282: Consider adding an explicit OSAC-2816 regression test that documents *why* this suite exists. Something like:

```python
def test_list_default_accepted_osac_2816_regression(self):
    """Regression: OSAC-2816 — list defaults were silently dropped."""
    defn = TemplateParameterDefinition(
        name="tags", title="Tags", description="Tags",
        type="list", required=False, default=["a", "b"],
    )
    param = TemplateParameter.from_definition(defn)
    assert param.default is not None
    assert param.default.value == ["a", "b"]
```

The current tests cover this implicitly through `test_collection_defaults`, but a named regression test makes the motivation discoverable for future contributors.

alosadagrande (commented):
Overall looks good — the `TestRealTemplateMetadata` parametrized test that parses all real `osac.yaml` files is the core value here and would have caught OSAC-2816. A few minor suggestions below, none are blockers.

tests/conftest.py
tzvatot: Replaced `sys.path.insert` with `pythonpath` in `[tool.pytest.ini_options]` and deleted the conftest. Done in 41609e1.

pyproject.toml
tzvatot: Switched to `>=` to match the existing style. The `==` pinning was from addressing a CodeRabbit comment, but consistency with the rest of the dev deps makes more sense. Done in 41609e1.

tests/unit/plugins/filter/test_find_template_roles.py
tzvatot: Added a `TestRegressions` class with explicit list and dict regression tests for OSAC-2816. Also added a dict variant since that was affected by the same bug. Done in 41609e1.

PR #455, opened 2026-07-29
OSAC-1992: read tier definitions and backend credentials from AAP extra_vars

wgordon17:
## Summary
- Replaces the `STORAGE_TIERS` env var and `storage-operations-ig` VAST admin credentials with `ansible_eda.event.storage_tier_definitions`/`storage_backend_connections` extra_vars populated by osac-operator's new Tier API integration, absorbing OSAC-1957's outstanding AAP-side AC
- Centralizes `qos_policy` derivation in the `storage_provider` role so every caller gets consistent naming without duplicating the logic
- Wraps AAP-discovered tier/backend capability-mismatch checks (unsupported protocol, ambiguous/unresolved/incomplete backend connections) in structured `set_stats` failure artifacts, additive scaffolding for a future osac-operator consumer to attribute job failures to a specific StorageTier/StorageBackend

## Summary by CodeRabbit

* **New Features**
  * Storage provisioning/teardown is now event-driven via `storage_tier_definitions` and per-`backend_id` `storage_backend_connections`.
  * Quotas use `quota_bytes`; QoS policy is always derived from the tier name.
  * Storage tier dispatch supports per-provider filtering, with stricter protocol validation and snapshot handling via live VolumeSnapshot CRD detection.
* **Bug Fixes**
  * Improved fail-fast validation and structured failure reporting for invalid inputs and credential/backend issues.
* **Documentation**
  * Updated provider docs, templates, and example payload/configs for the event-driven model.
* **Tests**
  * Expanded dispatcher and integration coverage for new validation, filtering, and credential/snapshot scenarios.

### DIFF
--- a/collections/ansible_collections/osac/config_as_code/roles/aap/templates/storage-operations-ig.j2
+++ b/collections/ansible_collections/osac/config_as_code/roles/aap/templates/storage-operations-ig.j2
@@ -34,11 +34,6 @@ spec:
         - configMapRef:
             name: storage-operations-ig
             optional: true
-      env:
-        - name: OSAC_STORAGE_CONFIG_NAMESPACE
-          valueFrom:
-            fieldRef:
-              fieldPath: metadata.namespace
   volumes:
     - name: kube-api-access
       projected:
--- a/collections/ansible_collections/osac/service/roles/storage_provider/tasks/_dispatch_provider.yaml
+++ b/collections/ansible_collections/osac/service/roles/storage_provider/tasks/_dispatch_provider.yaml
@@ -22,19 +22,7 @@
 - name: "Compute filtered tiers for current provider"
   ansible.builtin.set_fact:
     _computed_provider_tiers: >-
-      {{ (_storage_provider_tiers | selectattr('provider', 'equalto', _current_provider) | list)
-         | selectattr('name', 'in', _requested_tiers) | list
-         if (_requested_tiers is defined and _requested_tiers | length > 0)
-         else (_storage_provider_tiers | selectattr('provider', 'equalto', _current_provider) | list) }}
-
-- name: "Fail if requested tier filter produced no matches"
-  when: >-
-    _requested_tiers is defined and _requested_tiers | length > 0 and
-    _computed_provider_tiers | length == 0
-  ansible.builtin.fail:
-    msg: >-
-      None of the requested tiers ({{ _requested_tiers | join(', ') }}) match
-      provider '{{ _current_provider }}' tiers. Check storage_tier_definitions configuration.
+      {{ _storage_provider_tiers | selectattr('provider', 'equalto', _current_provider) | list }}
 
 - name: "Dispatch provider storage action with remote kubeconfig routing"
   module_defaults:
--- a/collections/ansible_collections/osac/service/roles/storage_provider/tasks/_dispatch_provider.yaml
+++ b/collections/ansible_collections/osac/service/roles/storage_provider/tasks/_dispatch_provider.yaml
@@ -7,7 +7,7 @@
 #   _dispatch_action   — task file to invoke (setup, ensure_storage_class, teardown_cluster_storage, teardown_backend)
 #
 # Uses from play scope:
-#   storage_provider_tiers              — full tier list (filtered here per provider)
+#   _storage_provider_tiers              — full tier list, qos_policy-defaulted (filtered here per provider)
 #   _cluster_name                        — optional target cluster name override (for volume
 #                                          naming); ensure_storage_class self-discovers it from
 #                                          the target cluster's Infrastructure object when unset
@@ -19,10 +19,10 @@
 - name: "Compute filtered tiers for current provider"
   ansible.builtin.set_fact:
     _computed_provider_tiers: >-
-      {{ (storage_provider_tiers | selectattr('provider', 'equalto', _current_provider) | list)
+      {{ (_storage_provider_tiers | selectattr('provider', 'equalto', _current_provider) | list)
          | selectattr('name', 'in', _requested_tiers) | list
          if (_requested_tiers is defined and _requested_tiers | length > 0)
-         else (storage_provider_tiers | selectattr('provider', 'equalto', _current_provider) | list) }}
+         else (_storage_provider_tiers | selectattr('provider', 'equalto', _current_provider) | list) }}
 
 - name: "Fail if requested tier filter produced no matches"
   when: >-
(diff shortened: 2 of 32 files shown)


wgordon17:
@coderabbitai review

wgordon17:
@coderabbitai review

playbook_osac_delete_tenant_cluster_storage.yml
```diff
@@ -37,25 +44,18 @@
...
-            fail_msg: >-
-              STORAGE_TIERS must be a JSON array of tier objects.
-      rescue:
-        - name: Fail on malformed STORAGE_TIERS JSON
-          ansible.builtin.fail:
-            msg: >-
-              STORAGE_TIERS env var contains invalid JSON.
-              Value: '{{ _storage_tiers_raw }}'.
-              Expected a JSON array of tier objects with name, protocol, and provider fields.
+    - name: Validate storage_tier_definitions is a non-empty array in the EDA event
```
zszabo-rh, line 47: I'm not sure about this is really necessary here.. missing tier definitions on creation is definitely a real misconfiguration but for teardown this may be too strict.
Not asking for a full change, just consider whether this should be warn + skip rather than fail when `storage_tier_definitions` is absent on the delete path.

zszabo-rh (commented):
One minor concern noted inline, otherwise the changes LGTM! (👍🏻ᴗ _ᴗ)👍🏻

playbook_osac_create_compute_instance.yml
```diff
@@ -71,9 +79,9 @@
       vars:
         storage_provider_action: ensure_storage_class
         storage_provider_tiers: "{{ _jit_storage_tiers }}"
+        storage_provider_backend_connections: "{{ ansible_eda.event.storage_backend_connections | default({}) }}"
         storage_provider_provisioning_target: vmaas
```
akshaynadkarni, line 83: nit:
Is this being used? Are there plans to use this in the future for caas, bmaas etc.?

collections/ansible_collections/osac/service/roles/storage_provider/tests/test.yml
akshaynadkarni, line 298: nit: Test comments jump from 7 to 9 to 11 (missing 8, 10). Likely from removed scenarios during iteration. Perhaps we can skip numbering.

tests/integration/run_tests.sh
akshaynadkarni, line 202: nit: The `STORAGE_TESTS` targets (`storage_provider_setup`, `storage_provider_ensure_sc`, etc.) are VAST-specific end-to-end: they hardcode provider: `vast`, assert `vast-*` resource names, and require the mock VMS server. When a second provider lands (e.g. PURE, LVMS), these names will be ambiguous.

Should we add a prefix or a `vast/` sub-directory?
WDYT?

collections/ansible_collections/osac/service/roles/storage_provider/tasks/_dispatch_provider.yaml
akshaynadkarni, line 38: nit: Today this dispatch loop is only exercised with a single provider (vast) in tests. Once LVMS lands (cc @zszabo-rh ), can we add a test with tiers spanning two providers? WDYT?

collections/ansible_collections/osac/templates/roles/vast_storage/tasks/ensure_storage_class.yaml
```diff
@@ -307,17 +310,20 @@
...
+      ansible.builtin.set_fact:
+        _vast_snapshot_crd_available: "{{ _vast_vsc_crd_check.resources | default([]) | length > 0 }}"
+
     - name: Create VolumeSnapshotClass per tier
-      when: >-
-        (storage_provider_snapshots_enabled | default(true)) and
-        (_vast_vsc_crd_check.resources | default([]) | length > 0)
+      when:
+        - _vast_snapshot_crd_available | bool
+        - storage_provider_snapshots_enabled | default(true) | bool
```
akshaynadkarni, line 326: question: `storage_provider_snapshots_enabled` defaults to true, meaning every `ensure_storage_class` call creates a `VolumeSnapshotClass` per tier alongside the StorageClass. Two questions:

1. Are there side effects to creating `VolumeSnapshotClasses` by default? 
e.g., does the VAST CSI driver need additional RBAC or a specific version to handle snapshot operations, and what happens if a user tries to create a `VolumeSnapshot` on a cluster where the driver doesn't support it?

2. This parameter isn't declared in `storage_provider/meta/argument_specs.yaml`, so consumers of the role won't know it exists unless they read the task file. Perhaps you can add it for completeness.

collections/ansible_collections/osac/config_as_code/roles/aap/templates/storage-operations-ig.j2
akshaynadkarni, line 33: question: After this PR the Secret has no credential fields left, but the `secretRef` is still required (no optional: true). The [compute-instance IG already marks it optional](https://github.com/osac-project/osac-aap/blob/f88bf3f61ab6ff41db0030c237d0a2d12c671a11/collections/ansible_collections/osac/config_as_code/roles/aap/templates/compute-instance-operations-ig.j2#L39-L45). Should this match the compute-instance IG?

I am not seeing a reason to keep it required, for now. WDYT?

wgordon17:
This PR has been migrated to the new osac mono-repo as https://github.com/osac-project/osac/pull/99, following Elior Erez's guidance in #wg-osac-eng (2026-07-31) that open PR authors should port their work to `osac-project/osac` now that osac-aap has merged upstream ([OSAC-1732](https://redhat.atlassian.net/browse/OSAC-1732)).

This is a migration, not an abandonment — the full 22-commit history and review thread remain here for reference.

**Review findings addressed in the new PR:**

@akshaynadkarni @zszabo-rh — here is the status of each substantive item from round 6:

1. **argument_specs entry** (`storage_provider_snapshots_enabled`) — **Fixed.** Added to `storage_provider/meta/argument_specs.yaml` (type bool, default true). The VolumeSnapshotClass CRD live-check already gates creation regardless of this flag, so no real side effect exists — this is a documentation-completeness fix only.

2. **secretRef optionality** (`storage-operations-ig.j2`) — **Fixed.** Marked `optional: true` to match `compute-instance-operations-ig.j2`. The Secret has no credential fields remaining (credentials now sourced via `storage_backend_connections` extra_vars from osac-operator), no consumer reads from it, and no fallback to broader-access credentials exists. An inline comment in the template references the credential-isolation decision.

3. **Teardown warn-vs-fail** (`playbook_osac_delete_tenant_cluster_storage.yml`) — **Retained as-is** after research. Warn+skip would let osac-operator mark cleanup as complete when the `storage_provider` dispatch never ran (absent tier data means no provider routing), orphaning cluster-side StorageClasses/VolumeSnapshotClasses/CSI Secrets. The hard fail forces reconciliation retry, which is correct since the only production case for absent tier data is a transient Tier API failure.

The 4 forward-looking nits (CaaS/BMaaS field usage, multi-provider test layout, multi-provider dispatch test, test-comment numbering) are carried into the new PR's description for continuity.

